PR #12, opened 2014-09-13
Support for VCSH

ariver:
Specifically, I set variables for GIT_DIR and the ATTRIBUTES file.
Then use those variables throughout script.
Updated comments and documentation a bit.
Hope it's all to your liking.

### DIFF
--- a/man/transcrypt.1.ronn
+++ b/man/transcrypt.1.ronn
@@ -72,8 +72,8 @@ were not given.
 
 Once a repository has been configured with transcrypt, you can transparently
 encrypt files by applying the "crypt" filter and diff to a pattern in the
-git attributes config. If that pattern matches a file in your repository,
-the file will be transparently encrypted once you stage and commit
+top-level _.gitattributes_ config. If that pattern matches a file in your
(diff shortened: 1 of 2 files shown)


elasticdog:
Hey @ariver, so I'm a terrible person and have just let this PR sit around. Initially I didn't quite grok what it was doing (or really why it would be beneficial), but I was investigating another request regarding bare repositories and this suddenly clicked. I've got a branch that basically does what you have here, but could use some help with testing via VCSH specifically.

Could you document some steps of how you would initialize a repository with VCSH and then use it with transcrypt? I found the upstream VCSH docs rather lacking. Thanks in advance.

ariver:
Well, **I've** been a terrible person and didn't notice you commented on my PR. =]
If you still want my input on this, I'd be happy to provide it.

elasticdog:
@ariver I could definitely use some usage examples as mentioned in order to figure out the workflow and how transcrypt should fit in with VCSH. I have a suspicion that the work to get VCSH functioning as expected will be beneficial to #22 as well.

PR #29, opened 2016-08-28
Do not alter output with repeated filter commands

### LINKED ISSUE #27: Diff problems when an empty file is marked for encryption
Steps to reproduce
  1. Create git repo
  1. Initialize Transcrypt
  1. Add Transcrypt rule to `.gitattributes`
    - `empty-file filter=crypt diff=crypt`
  1. Add/Commit the `.gitattributes` file
  1. Create the `empty-file`
    - `touch empty-file`
  1. Add the `empty-file`

At this point, if you run `git diff` there will be an empty diff, instead of no diff. If you run `git diff --exit-code` and then check the exit code `echo $?`, you'll see it exits with a `1` (indicating a difference), instead of a `0` (indicating no difference). If you reproduce these steps except for initializing transcrypt and configuring the rule for the empty file, `git diff --exit-code` will exit with a `0`.

From the user's perspective there's no indication what the problem is and it interferes with the ability to run any commands that depend on a clean working tree, eg `git rebase`.

elasticdog:
Per the `gitattributes` man page, the filter command behavior should be a bit more strict:

> For best results, clean should not alter its output further if it is run
> twice ("clean->clean" should be equivalent to "clean"), and multiple smudge
> commands should not alter clean's output ("smudge->smudge->clean" should be
> equivalent to "clean"). See the section on merging below.

This does add some complexity surrounding the capturing of STDIN, but I think the tempfiles are okay. I'm going to keep testing with this branch for a bit before I merge to see if there are any unintended consequences.

This also allows us to handle the empty files more gracefully (by skipping them explicitly), so it will resolve #27.

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -211,10 +211,13 @@ save_helper_scripts() {
 	cat <<-'EOF' > "${REPO}/.git/crypt/clean"
 		#!/usr/bin/env bash
 		filename=$1
-		cipher=$(git config --get --local transcrypt.cipher)
-		password=$(git config --get --local transcrypt.password)
-		salt=$(openssl dgst -hmac "${filename}:${password}" -sha256 "$filename" | tail -c 16)
-		ENC_PASS=$password openssl enc -$cipher -pass env:ENC_PASS -e -a -S "$salt"
+		# ignore empty files
+		if [[ -s $filename ]]; then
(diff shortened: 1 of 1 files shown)


elasticdog:
Been running through a lot of testing, and the behavior seems great. I haven't noticed any issues, so I'm going to go ahead and merge this.

PR #30, opened 2016-08-30
Generalize handling of repository path references

elasticdog:
Cleaning up the hard-coded references to certain repository paths not only reduces duplication, but allows us to support fancier workflows. By generalizing the `$GIT_DIR` and `$GIT_ATTRIBUTES` paths, we can now:

* configure bare repositories
* configure "fake bare" repositories for use through [vcsh](https://github.com/RichiH/vcsh)
* configure multiple worktrees via [git-workflow](https://github.com/blog/2042-git-2-5-including-multiple-worktrees-and-triangular-workflows)
* support unencrypted archive exporting via [git-archive](https://git-scm.com/docs/git-archive)

The trickiest part of all this was the need for converting relative paths into absolute paths in a reliable manner. External programs exist to do this, but they're not commonly part of the base OS installs for all the platforms I'd like to support. I needed a Bash-only solution...

I ended up writing a `realpath()` function, and it has proven robust in my testing across various workflows thus far, but I suspect that it's the weak link in these changes. To help understand what's going on, transcrypt now displays some key repository metadata as part of the `--display` command.

ping @tvaughan @ariver @KenRud any chance you fellas could test this branch with your desired workflow to see if things work as expected outside of my limited testing environments?

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -194,6 +194,7 @@ confirm_configuration() {
 	local answer
 
 	printf '\nThe following configuration will be saved:\n\n'
+	[[ $REPO ]] && printf '  GIT_WORK_TREE:  %s\n' "$REPO"
 	printf '  GIT_DIR:        %s\n' "$GIT_DIR"
 	printf '  GIT_ATTRIBUTES: %s\n\n' "$GIT_ATTRIBUTES"
 	printf '  CIPHER:   %s\n' "$cipher"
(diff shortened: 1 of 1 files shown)


elasticdog:
I've been playing around with this for the past couple of days and am happy enough to put it into the next release. I'm going to merge, but please do let me know if any of you experience issues with transcrypt's handling of these various repository paths.

PR #33, opened 2016-12-31
small changes to make transcrypt run in termux

fbartels:
Hi Aaron,

I played around a bit with your script inside of https://termux.com/. At least on my Android device there is no /tmp, which will make the mktemp command fail. I also had to manually install column after the first execution, since it absence was not catched.

This pull request fixes both issues.

Regards

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -110,7 +110,7 @@ run_safety_checks() {
 	fi
 
 	# check for dependencies
-	for cmd in {grep,mktemp,openssl,sed,tee}; do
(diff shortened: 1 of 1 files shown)


elasticdog:
Good catch on the `column` check, and I think I originally had the mktemp file template because I didn't read the man page on OS X closely enough and assumed it was required. Looks like without a template, it does default to more sane behavior everywhere.

I went ahead and merged this into master and pushed another commit to update the other `mktemp` usage. Thank you for the contribution!

fbartels:
Thanks for your quicke merge!

Issue #40, opened 2017-04-28
Cannot decrypt repositories with OpenSSL 1.1.0 and later, that were created with earlier versions of OpenSSL


ircecho:
The reason is that [default digest for the dgst and enc commands from MD5 to sha256](https://www.openssl.org/news/cl110.txt).

When setting the digest to md5, the files can be decrypted properly again.

lynchem:
Could you share your change @ircecho  ? We have the same problem.

ircecho:
You have to change the following lines in the files in the folder ".git/crypt":

```
clean: ENC_PASS=$password openssl enc -$cipher -md MD5 -pass env:ENC_PASS -e -a -S "$salt" -in "$tempfile"
smudge: tee "$tempfile" | ENC_PASS=$password openssl enc -$cipher -md MD5 -pass env:ENC_PASS -d -a 2> /dev/null || cat "$tempfile"
textconv:  ENC_PASS=$password openssl enc -$cipher -md MD5 -pass env:ENC_PASS -d -a -in "$filename" 2> /dev/null || cat "$filename"
```

See the added **-md MD5**.

lynchem:
Thanks for that @ircecho . All working nicely.

thorntonrose:
This fix worked for me on a Debian 8 system. Thanks. When will this fix be released?

elasticdog:
This has been resolved by #41 and I'm working on cutting a new release as we speak.

PR #41, opened 2017-08-17
Specify MD5 as the digest to create a key from the passphrase.

damaestro:
Address https://github.com/elasticdog/transcrypt/issues/40 by setting the digest to MD5 for this operation. The question to be raised is should there be automatic migration to a different digest in the later versions of openssl? Is this being hardcoded a *bad* thing?

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -284,7 +284,7 @@ save_helper_scripts() {
 		    cipher=$(git config --get --local transcrypt.cipher)
 		    password=$(git config --get --local transcrypt.password)
 		    salt=$(openssl dgst -hmac "${filename}:${password}" -sha256 "$filename" | tail -c 16)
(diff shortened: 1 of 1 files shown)


elasticdog:
Hey @damaestro!

There's no automated migration in the upstream libraries, just a [change in the default](https://github.com/openssl/openssl/commit/9e8b6f042749ded556380227c9f2db7ffad9a3aa). As long as it's hard-coded in the clean/smudge filter scripts, we can also change the transcrypt default settings to match upstream, but I'll probably do that over a period of time after this PR is merged and a new release has been cut, in order to provide time for people to migrate to the explicit versions of the script and not break things moving forward. I'll think about if it would make sense to also add an option flag for setting the digest type at that time.

Thanks for the PR, I've already pushed your changes to the master branch and will cut a new release shortly.

Issue #39, opened 2017-04-28
"openssl list-cipher-commands" does not work with OpenSSL 1.1.0e; should be "openssl list -cipher-commands"


ircecho:

ircecho:
This problem is only present in the later OpenSSL versions, mine is 

```
OpenSSL 1.1.0e  16 Feb 2017
```

dfee:
Specifically, here is the update that needs to be made:

```bash
validate_cipher() {
	local supported=$(openssl list -cipher-commands | tr -s ' ' '\n' | grep --line-regexp "$cipher")
	if [[ ! $supported ]]; then
		if [[ $interactive ]]; then
			printf '"%s" is not a valid cipher; choose one of the following:\n\n' "$cipher"
			openssl list -cipher-commands | column -c 80
			printf '\n'
			unset cipher
		else
			die 1 '"%s" is not a valid cipher; see `openssl list -cipher-commands`' "$cipher"
		fi
	fi
}
```

There are two changes:

1. `openssl list -cipher-commands` as mentioned above
2. `openssl list -cipher-commands | tr -s ' ' '\n' | grep --line-regexp "$cipher"` note the `tr -s` to turn the tabular data into a single column list.

If my bash-fu were better, I'd issue a PR, but I can't quite figure out how to do a conditional on the semver of openssl. i.e. `openssl version | awk '{print $2}'` returns the version... and we know that 1.1 is where the change broke, but I'm not quite there in bash land for parsing major / minor / patch and doing the comparison. Hopefully this is a good starting point for the author.

For now, I'm just adding it to my repo's /scripts dir.

jackrh:
This issue popped up for me when I tried to decrypt a new cloned repository using the provided transcrypt command from the original repo. Applying dfee's validate_cipher update fixed it for me.

cdimitroulas:
I have made a PR from @thomas-hilaire's fork which has the fix for this - not sure why he didn't make it himself.

elasticdog:
This has been resolved by #48 and a7e8aef. I'll be cutting a new release shortly, but feel free to try out the code in the _master_ branch.

PR #47, opened 2018-04-16
Update OpenSSL command to openssl list -cipher-commands

cdimitroulas:
see https://github.com/elasticdog/transcrypt/issues/39


elasticdog:
Resolved by #48 and a7e8aef

PR #48, opened 2018-05-02
Fix OpenSSL v1.1+ incompatibility

Aramgutang:
In OpenSSL 1.1, the equivalent of `list-cipher-commands` is now `list -cipher-commands`, and the former is no longer supported.

This fix maintains backwards compatibility by checking which of the commands succeed first, and use the working one in the subsequent code.


cdimitroulas:
@elasticdog can you please have a look at this? It would be really great if you could review and merge soon if this is ok :)

PR #50, opened 2018-12-19
make sure to use system's sed on OSX

andreineculau:
or else risk using a GNU sed from `brew install gnu-sed --with-default-names`.
Alternatively one can check for gsed on OSX and use that instead.


elasticdog:
Nice catch on the potential environmental incompatibility...really appreciate you taking the time to push this fix. Merging.

PR #51, opened 2018-12-19
add .editorconfig

andreineculau:
see https://editorconfig.org


elasticdog:
Apologies on taking forever to respond...this look perfect, and I've been getting into the habit of adding an EditorConfig for my other repos as well. Thanks for the PR!

PR #52, opened 2018-12-19
silence "fatal: no such section: <name>" messages

andreineculau:
stemming from `git config --remove-section` commands

git 2.18+ (Q2 2018) will remove empty sections on its own
as per https://stackoverflow.com/a/50274206/465684


elasticdog:
It has been a hot minute since I've touched transcrypt and had noticed these messages when working through the backlog of PRs...cheers for saving me the trouble of diving into where those messages were coming from.

PR #53, opened 2018-12-19
use strict bash

andreineculau:
see http://redsymbol.net/articles/unofficial-bash-strict-mode/

PR on top of #52

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -373,15 +373,15 @@ display_configuration() {
 
 # remove transcrypt-related settings from the repository's git config
 clean_gitconfig() {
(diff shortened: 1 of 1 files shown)


elasticdog:
I have historically been averse to using Bash "strict" mode due to all the unintuitive [edge cases](https://mywiki.wooledge.org/BashFAQ/105), so I was admittedly averse to this PR when I saw the subject, but I was pleasantly surprised by your changes. Very cleanly done, and I do think it's a nice improvement! Thanks for all the contributions, and for pushing my own thinking on the matter.

Issue #55, opened 2019-03-19
Suppress warnings for OpenSSL 1.1.1


dirkhas:
Using OpenSSL 1.1.1 (in my case, in Git for Windows, which uses `OpenSSL 1.1.1a  20 Nov 2018`), the `clean` script constantly leads to `git` commands printing out the following warning:
```
*** WARNING : deprecated key derivation used.
Using -iter or -pbkdf2 would be better.
```

dirkhas:
I'll create a pull request to fix it.

ricardo0624:
in transcrypt 1.1.0, I still get this warning,it recommend 'Using -iter or -pbkdf2 would be better.'

elasticdog:
This will be a bit tricky to address, as OpenSSL 1.1.1 was released in Sep 2018 and I want to make sure that older versions also work properly. The situation is further complicated by macOS using LibreSSL 2.6.5 these days, which does not support the `-pbkdf2` option, as far as I know. I'll have to do a bit more research here to know how far back the `-iter` option has been around in both projects to see if it makes sense to update the default behavior and introduce a breaking change.

I don't think suppressing the warning is the best way forward in this case since it did bubble up a legit deprecation that will bite us eventually.

Erotemic:
Perhaps it is possible to do a version check and then actually use -pbkdf2 and -iter when calling openssl, as long as you have at least 1.1.1?

andreineculau:
I'm wondering if anyone has a clear path going forward.

Both `-pbkdf2` and `-iter` have landed in 1.1.1 afaik and libressl 2.9.1, so in order to support previous versions, you need to check for the support of the flags. But that is not enough because it should be an active decision of transcrypt not to support openssl@<1.1.1/libressl@<2.9.1, or of the transcrypt user not to support openssl@<1.1.1/libressl@<2.9.1, otherwise you risk weird behavior on two different machines, one with, and the other without these flags.

If someone decides, I can file a PR to test out the decision, no problem.

jmurty:
I think the best approach for handling this situation would be for transcrypt to switch to using the more secure `-pbkdf2` and `-iter` options by default, but in a way that minimises breaking changes as much as possible.

Here are some rough ideas on how to do this, and a longwinded (sorry) brain-dump about how it could all work.

### Additional features

[...]

PR #57, opened 2019-03-19
#56 Use 16 hex characters in salt

### LINKED ISSUE #56: Openssl 1.1.1 salt warnings / Git for Windows interoperability
In [Git for Windows](https://git-scm.com/download/win), currently in version 2.21.0 using Bash 4.4.23 and OpenSSL 1.1.1a, the salt generation in the `clean` script causes two issues:

a) The following warning is issued by OpenSSL everytime the `openssl enc` part of the `clean` script is called: `hex string is too short, padding with zero bytes to length`
b) salt generation differs between *nix and Windows due to line-endings.

Both issues have the same solution: make the salt consist of 16 hex characters not including *any* line-ending characters. Pull request on the way.

dirkhas:
- The current salt uses 15 hex characters and the new-line character on
*nix, and 14 hex characters and carriage-return plus new-line characters
on Windows. Stripping both characters in salt-generation makes
the code cross-platform and eliminates the OpenSSL warning that would
otherwise be printed.
- WARNING: This requires re-encrypting secrets because salts change!
- Fixes #56

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -292,7 +292,7 @@ save_helper_scripts() {
 		  else
 		    cipher=$(git config --get --local transcrypt.cipher)
 		    password=$(git config --get --local transcrypt.password)
-		    salt=$(openssl dgst -hmac "${filename}:${password}" -sha256 "$filename" | tail -c 16)
+		    salt=$(openssl dgst -hmac "${filename}:${password}" -sha256 "$filename" | tr -d '\r\n' | tail -c 16)
 		    ENC_PASS=$password openssl enc -$cipher -md MD5 -pass env:ENC_PASS -e -a -S "$salt" -in "$tempfile"
 		  fi
 		fi


dirkhas:
Because #55 already suppresses warnings, a way to make a backwards-compatible change that would fix the cross-platform issues described in #56 (i.e. make the salt-derivation work identical on both *nix and Windows) would be to substitute
`tr -d '\r'`
for
`tr -d '\r\n'`
With no '\r' to remove on *nix, the change only has effect on Windows.

leclairmael:
@dirkhas Thanks for the PR, looks good. Hopefully it gets released soon.

ricardo0624:
why not tail -c 17 ?

leclairmael:
@ricardo0624 The salt needs to be exactly 16 characters.

ricardo0624:
@leclairmael tail -c 17 is exactly 16 characters. op ,tail -c 16 is just 15 characters.

dirkhas:
> @leclairmael tail -c 17 is exactly 16 characters. op ,tail -c 16 is just 15 characters.

`tail -c 17` is exactly 17 characters and `tail -c 16` is exactly 16 characters, but if you feed something ending in a new-line -- for example anything from `echo` -- the last character is a new-line. So it *looks like* there is one less character.

Try for example
`echo "1234567890abcdef" | tail -c 16 | wc` and notice than 16 characters are counted. Also notice that
`echo "1234567890abcdef" | tail -c 16` *doesn't print the '1'* because echo appends a new-line character so 17 characters are fed to `tail` but only the last 16 should be printed.

elasticdog:
@dirkhas Sorry for the radio silence on all these issues...thank you for noticing the inconsistency and for providing the fix! What's the proposed workflow for re-encrypting in this case? Once I can sort out easy directions for people to migrate, I'll get this merged and will cut a new release once I've gone through the rest of the backlog.

elasticdog:
I guess this would be it after updating transcrypt:

```
$ git add -- $(transcrypt --list)
$ git commit --message="Re-encrypt files protected by transcrypt using new salt value"
```

elasticdog:
Actually, it's not quite that simple since the repository's configured clean filter script will have to be updated, and transcrypt won't do that if the repo is dirty. Need to think about this a bit more, as I think this is the first real backward incompatible change the project has needed. If you have a working set of steps, please do let me know...

elasticdog:
For posterity, flushing the credentials first and then re-configuring the repo before adding the files does the trick:

```
$ transcrypt --display
The current repository was configured using transcrypt version 1.1.0
and has the following configuration:

  GIT_WORK_TREE:  /home/elasticdog/src/transcrypt
  GIT_DIR:        /home/elasticdog/src/transcrypt/.git
  GIT_ATTRIBUTES: /home/elasticdog/src/transcrypt/.gitattributes

  CIPHER:   aes-256-cbc
  PASSWORD: correct horse battery staple

Copy and paste the following command to initialize a cloned repository:

  transcrypt -c aes-256-cbc -p 'correct horse battery staple'
$ transcrypt --flush-credentials
$ transcrypt -c aes-256-cbc -p 'correct horse battery staple'
$ git add -- $(transcrypt --list)
$ git commit --message="Re-encrypt files protected by transcrypt using new salt value"
```

PR #58, opened 2019-03-19
#55 Suppress warnings during clean

### LINKED ISSUE #55: Suppress warnings for OpenSSL 1.1.1

dirkhas:
- OpenSSL 1.1.1 and later issues warnings with the current clean script.
These have now been disabled by piping stderr to /dev/null similar to
how it's done in smudge and textconv.
- Fixes #55


elasticdog:
I left a comment in https://github.com/elasticdog/transcrypt/issues/55#issuecomment-496022602 about why I don't think suppressing the warning is the right move in this case...I'm going to close out this PR while looking into the possibility of changing the default message digest configuration, which would also remove the warning by addressing the root cause.

Issue #25, opened 2015-12-28
Multiple keys, same repo


charliewolf:
Oh, hi there. 

Is there any way to use multiple transcrypt keys in the same repo? i.e. encrypt some files with key A and some with key B. 

I can try to write some code if you point me in the right direction. Let me know! Thanks.

elasticdog:
Hey Charlie!

So, it would take quite a bit of tweaking. Basically you'd need a new dedicated section in the config file where it would store the credentials and filter definitions, and then also separate clean/smudge filter scripts that pull in those credentials. Then you'd select which crypt filter to apply to a particular pattern of files in the *.gitattributes* file. That said, the UI would also need updating to manage multiple keys and also pass the safety checks accordingly, so overall it would be a complicated undertaking.

elasticdog:
@charliewolf! Did you end up implementing something for this, or work around it another way? Honestly, I have been neglecting transcrypt as I haven't been using it (personally or professionally) for long time.

charliewolf:
@elasticdog I haven't been using `transcrypt` either. Mainly have used `git-crypt` when I've needed to encrypt things in a repo

cdimitroulas:
@elasticdog what are you using instead of `transcrypt`? I was about to use this library for encrypting some variables in a repo for out team but noticed your comment saying you aren't using this personally or professionally which has kind of put me off...

elasticdog:
@cdimitroulas rather than have secrets stored in the repo itself, we've been using [HashiCorp's Vault](https://www.vaultproject.io/) as a dedicated, highly-available secret store and have various tooling/workflows to pull things out as needed. Lots of advantages to using something like Vault, but certainly more work and infrastructure to manage.

PR #63, opened 2019-06-07
PROPOSAL: Add support for multiple passphrase "contexts" in one repo

jmurty:
Hi, thanks for your great work on *transcrypt*! We have found it really useful.

We are considering using *transcrypt* for a scenario it wasn't designed for, where we want to have multiple "contexts" of secret file within the one repository with a different passphrase for each one. 
Basically, we have some files in a repo that should be encrypted but available to user group A, and another set of files that should only be available to user group B. 

This PR contains a **very** rough but working proof of concept with *transcrypt* updated to support this use case, where different security "contexts" can be specified by setting a `CONTEXT` environment variable before running the script.

**If you are at all interested in adding such a feature please let me know here, along with any thoughts about how you would prefer it to be built.** We can work on improving the implementation if you think it might get included in the official version later?

The implementation here so far is a quick and dirty hack that duplicates all the git hook scripts for each named context, so that each one knows about its particular settings.

Despite being an awful hack it shows that the process works fairly well, with just those files for which the user has set the appropriate context & passphrase decrypted, and the rest left as they were.

Example usage:

- Choose a name for a non-default encryption "context" e.g. "super-secret"
- Init transcrypt in default and "super-secret" contexts:
  ```bash
  $ transcrypt
  # follow prompts

  $ CONTEXT=-super-secret transcrypt
  # follow prompts, note leading dash (-)
  ```
- Set patterns in `.gitattributes` as usual for default decryption, and with the suffix "-super-secret" for that additional context:
    ```
    # Contents of .gitattributes
    secret.txt filter=crypt diff=crypt
    super-secret.txt filter=crypt-super-secret diff=crypt-super-secret
    ```
- Write your secret files, `git add` them, and commit.
- Note the different passphrases for default and additional contexts
  ```bash
  $ transcrypt --display

  $ CONTEXT=-super-secret transcrypt --display
  ```
- Clone repo elsewhere, and confirm that you can or cannot see file contents as you apply some combination of the init commands as printed by the display commands above.

### DIFF
--- a/.github/workflows/run-bats-core-tests.yml
+++ b/.github/workflows/run-bats-core-tests.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-16.04, ubuntu-18.04]
+        os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
 
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
--- a/README.md
+++ b/README.md
@@ -322,8 +322,21 @@ Please use:
 Tests are written using [bats-core](https://github.com/bats-core/bats-core)
 version of "Bash Automated Testing System" and stored in the _tests/_ directory.
 
-To run the tests:
+To run the tests locally:
 
 - [install bats-core](https://github.com/bats-core/bats-core#installation)
 - run all tests with: `bats tests/`
 - run an individual test with e.g: `./tests/test_help.bats`
+
+To run the tests in Docker:
+
+- install Docker
+- `cd tests/`
+- check available test targets (Docker services): `docker-compose ps`
+- build images for all test targets: `docker-compose build`
+  - or for a specific test target: `docker-compose build ubuntu-20.04`
+- run tests on all targets (non-zero exit codes means fail): `docker-compose up`
+  - or run tests on a specific target: `docker-compose run --rm ubuntu-20.04`
+- to manually run and debug tests:
+  - run shell in a specfic target: `docker-compose run --rm ubuntu-20.04`
+  - run tests selectively with something like: `bats tests/test_init.bats -t`
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ by running the `--display` command line option:
     The current repository was configured using transcrypt v0.2.0
     and has the following configuration:
 
+      CONTEXT:  default
       CIPHER:   aes-256-cbc
       PASSWORD: correct horse battery staple
 
--- a/man/transcrypt.1.ronn
+++ b/man/transcrypt.1.ronn
@@ -65,6 +65,15 @@ The transcrypt source code and full documentation may be downloaded from
   * `-i`, `--import-gpg`=<file>:
     import the password and cipher from a gpg encrypted file
 
+  * `-C`, `--context`=<context_name>
+    name for a context that can use a different passphrase and cipher
+    from the 'default' context; use this advanced option, to permit
+    encrypting different files with different passphrases
+
+  * `--list-contexts`
+    list all contexts configured in the repository, and warn about
+    incompletely configured contexts.
+
   * `-v`, `--version`:
     print the version information
 
@@ -101,6 +110,29 @@ force a checkout of any existing encrypted files in order to decrypt them.
 If the origin repository has just rekeyed, all clones should flush their
 transcrypt credentials, fetch and merge the new encrypted files via Git, and
 then re-configure transcrypt with the new credentials.
+
+You can use context names to encrypt files for different audiences with
+different passwords, such as normal users versus admins. The 'default'
+context applies unless you set a context name.
+To add a context, initialise transcrypt with the context's name and add
+a .gitattributes pattern including 'crypt-context=<name>'. For example:
+
+    # Initialise a new context "admin", and set a different password
+    $ transcrypt --context=admin
+
+    # Add a pattern with 'crypt-context=admin' to .gitattributes
+    $ echo >> .gitattributes \\
+    'admin.txt filter=crypt diff=crypt merge=crypt crypt-context=admin'
+
+    # Add and commit your secret and .gitattribute files
+    $ git add .gitattributes admin.txt
+    $ git commit -m "Add encrypted admin-only file in 'admin' context"
+
+    # List all contexts
+    $ transcrypt --list-contexts
+
+    # Display a context's cipher and password
+    $ transcrypt --context=admin --display
 
 ## AUTHOR
 
--- a/tests/Dockerfile.ubuntu
+++ b/tests/Dockerfile.ubuntu
@@ -2,7 +2,10 @@ ARG version
 
 FROM ubuntu:$version
 
-# Install git
+# Install git, but PPA tool first so we can get newer git on 16.04
+RUN apt-get update \
+    && apt-get install -y software-properties-common \
+    && add-apt-repository ppa:git-core/ppa
 RUN apt-get update \
     && apt-get install -y git bsdmainutils
 
--- a/tests/Dockerfile.ubuntu
+++ b/tests/Dockerfile.ubuntu
@@ -1,0 +1,18 @@
+ARG version
+
+FROM ubuntu:$version
+
+# Install git
+RUN apt-get update \
+    && apt-get install -y git bsdmainutils
+
+# Install and set up bats-core
+RUN git clone https://github.com/bats-core/bats-core.git /tmp/bats-core-repo \
+    && mkdir -p /tmp/bats-core \
+    && bash /tmp/bats-core-repo/install.sh /tmp/bats-core
+RUN ln -s /tmp/bats-core/bin/bats /usr/local/bin/bats
+
+# Need to configure git globally to avoid errors when we create and use
+# temporary test-specific Git repos fatal: `empty ident name`
+RUN git config --global user.name github-actions-user \
+    && git config --global user.email github-actions-user@example.com
--- a/tests/_test_helper.bash
+++ b/tests/_test_helper.bash
@@ -66,3 +66,7 @@ function teardown {
   cleanup_all
   popd
 }
+
+function check_repo_is_clean {
+  git diff-index --quiet HEAD --
+}
--- a/tests/_test_helper.bash
+++ b/tests/_test_helper.bash
@@ -40,10 +40,16 @@ function init_transcrypt {
 function encrypt_named_file {
   filename="$1"
   content=$2
+  context=${3:-default}
   if [ "$content" ]; then
     echo "$content" > "$filename"
   fi
-  echo "\"$filename\" filter=crypt diff=crypt merge=crypt" >> .gitattributes
+  if [ "$context" != "default" ]; then
+    context_crypt="$context-crypt"
+  else
+    context_crypt="crypt"
+  fi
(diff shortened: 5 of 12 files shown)


elasticdog:
Hey @jmurty! I'd certainly be open to coordinating with you to add this feature...you're actually not the first person to suggest it (see https://github.com/elasticdog/transcrypt/issues/25), so I'm sure it would be useful to others.

As far as the user experience, I wonder if using actual option flags to the script would make it more discoverable compared to introducing the new mechanism of environment variables. Using options would also allow us to add something like `--list-contexts`, but I also like the idea of everything being backward compatible and using a default context without any extra effort for users that wouldn't need them. What do you think?

> Note, I've recently pushed some automated reformatting commits (using [shfmt](https://github.com/mvdan/sh)), so it looks like you've got a few merge conflicts now with the current implementation.

jmurty:
Hi @elasticdog thanks for the feedback and the pointer to #25. Your quick summary there about required changes lined up pretty well with my experience: there were quite a few changes needed, all fairly small but spread out.

I agree that proper option flags is the right way to go. The code in this PR so far is full of quick and dirty hacks to show how multiple "contexts" *could* work, but definitely isn't how I think they *should* work. As such, I'm not worried about the merge conflicts as I'd start over anyway, working towards a cleaner and more robust implementation before I'd consider it a real candidate to be merged.

I will spell out in more detail below how I think it should work and some questions I have. If you could let me know if I'm way off track or what you think in general that would be great. I will aim to build a better implementation along these lines.

You should also be aware that I don't have a huge amount of bash scripting experience so will probably make some rookie mistakes along the way, keep an eye out for those.

## Design sketch for multiple password "contexts"

Naming:
- I'm not sure that "contexts" is the best name for this feature, is there a better name that would more clearly express what it does? "Levels", "Groups", ... ?

Usage:
- ensure everything works as before for users who do not use extra contexts. Users shouldn't have to know or care about this feature unless they specifically need it
- support a new optional `--context [context-name]` option flag that can be combined with all relevant other options
  - the context name must comply with some rules so we can safely use it within settings files and scripts. Probably something like alpha-numeric characters only, case-insensitive (force to lower-case internally)
  - when used alone or with `--cipher` and/or `--password` will init a new encryption "context" with its own section in settings
  - `--context` would cause the script to target a specific context when combined with many other options: `--display`, `--rekey`, `--flush-credentials`, `--uninstall`, `--list`, `--export-gpg`, `--import-gpg`
  - when multiple contexts are configured and the `--context` option is not provided, other options would process or print info about *all* existing contexts. E.g. `--display` would list all contexts, `--list` would list files for each context in turn, etc
    - this might be too ambitious? A simpler alternative would be to print a reminder to users that they have multiple contexts, e.g. if a user runs `transcript --rekey` the output might say something like: `"Files in the default context have been rekeyed. Files in [2] other contexts were not affected"`
- add documentation about all of the above

Implementation:

- handle just the `--context` option at top of script to adjust various constants to be context-specific, similar to 1ab83c8b but using an option flag instead of environment variable
- update helper scripts to recognise and handle non-default context names somehow, so unlike 1ab83c8b just one copy of each helper script is needed but it behaves differently for different contexts
  - I'd probably re-use the same environment variables already set at the top of the script for this
  - When writing the filter settings for a context, adjust the commands to pass through any context-specific data needed by the helper scripts

jmurty:
Hi @elasticdog I have made a start on the cleaned-up design I outlined above. I'd appreciate any feedback you have.

acro5piano:
any updates?

jmurty:
Hi @acro5piano I can try to resurrect this PR and bring it up-to-date with the latest changes on master. Would you be able and willing to test the contexts feature if I can get it working again?

acro5piano:
Yes sure. I would like to try the multiple passphrase feature. Thanks.

jmurty:
Hi @acro5piano I think I have gotten this PR branch into decent shape and a new, moderately comprehensive, unit test suite for contexts is now passing.

Can you  please check out the PR branch and see if it works for you? See the [basic instructions here](https://github.com/elasticdog/transcrypt/blob/bc220b4bd465a902eef047b5db5d5cc6f368be9d/transcrypt#L1089-L1107) for hints on how to use contexts.

jmurty:
Hi @acro5piano I think I have gotten this PR branch into decent shape and a new, moderately comprehensive, unit test suite for contexts is now passing.

Can you  please check out the PR branch and see if it works for you? See the [basic instructions here](https://github.com/elasticdog/transcrypt/blob/07a878d3129261a38f4d6355b6310db7f0b93e7c/transcrypt#L1089-L1107) for hints on how to use contexts.

jmurty:
Hi @acro5piano I have greatly cleaned up and improved the contexts implementation, so if you are testing please try this latest version. The setup instructions have changed, you now use a new `crypt-context` attribute in *.gitattributes* to match patterns to a context. See [the updated instructions](https://github.com/elasticdog/transcrypt/blob/1f93c02d9c8f143ff9972bb93bd204a30801a3a5/transcrypt#L1083-L1099).

jmurty:
Hi @elasticdog I have revisited this contexts work and greatly cleaned it up. I would appreciate any feedback or thoughts you have about it.

The commit trail is a mess, but if you check the final changes I don't think they are too disruptive.

Key points:
- extra contexts are defined by init-ing them with `transcrypt --context=my-context` and assigning corresponding file patterns in *.gitattributes* with the `crypt-context=my-context` attribute
- you list your contexts, and see misconfiguration hints when appropriate, with `transcrypt --list-contexts`
- list all crypt-ed files with `git ls-crypt` or just those in a specific context with `git ls-crypt-my-context` or `git ls-crypt-default` (default/no context)
- the smudge/clean/textconv scripts now check for the `crypt-context` attribute, which might affect performance
- the contexts features are reasonably covered by tests
- there are some unrelated cleanup changes to the merge script in this PR.

acro5piano:
@jmurty 
Awesome work. I'll try it soon.

acro5piano:
@jmurty 
Hi, I've tried your version and it's working precisely. Thank you for your fantastic work!

One question, what does `ts` mean in the instructions?
I think more concrete name is comprehensive for those who use this feature for the first time, such as:

- `admin`
- `superuser`
- `staging` 
- `production`

jmurty:
Hi @acro5piano that's great! Thanks for testing and confirming it works for you.

You're right that the `ts` example name is bad, I only used it to try to squeeze the *.gitattributes* `echo` command in the help text within 79 characters. I will try to find a better solution than using a short, but bad example name.

jmurty:
Hi @acro5piano I have improved the help text for contexts, along with a few bug fixes. Is this text better now? https://github.com/elasticdog/transcrypt/blob/b2db9284d1d079b83acb74e5890c54b01e8fe0bf/transcrypt#L1085-L1103

acro5piano:
@jmurty Thank you for the improve! Yes I think the new version is much comprehensive.

elasticdog:
Is the plan to have this be merged after the 2.1.0 release work, or would it be a part of it?

jmurty:
@elasticdog Let's leave the contexts work to settle instead of rushing it into the 2.1.0 release, since I haven't used it at all myself day-to-day and I'm still finding rough edges and bugs.

I'll merge it to `master` after the next release, and when it's a little more tested.

yjarrar:
Is there a reason this never got merged? just new conflicts since last approved?

jmurty:
Oops, no good reason it never got merged, it just fell off my radar after I no longer needed it. I'll put it back on my radar and see if I can get it back on track, though be warned I have little time nowadays.

Have you tried out the unmerged version @yjarrar and, if so, any feedback?

yjarrar:
Was hoping to see it merged before using, but I will start using it now and let you know how it goes.

jmurty:
Hi @yjarrar I wouldn't recommend using this old contexts branch long term, but if you could try it out briefly and confirm whether it works, makes sense, and is reasonably usable, that would be very helpful.

acro5piano:
I actually use this branch and it works well. I also hope this PR is merged into the main branch.

benclanet:
Hi 👋🏻 
Awesome work!
Is there a plan to merge this feature? This is exactly what we'd need!
Let me know if I can help 👍🏻

jmurty:
Thanks for the feedback everyone!

I have put some time today into bringing the branch with the "contexts" feature up-to-date in preparation to test, polish, and eventually merge this feature to the `main` branch.

I put the updates in a new branch to avoid the risk of breaking anyone using the original branch, and to have a place for this work directly in this repo rather than a fork. Please see the `multiple-passphrases-in-one-repo-wip` branch in this repository, and the related PR https://github.com/elasticdog/transcrypt/pull/145

However, please keep comments and feedback in this thread as much as possible to avoid forking the conversation.

jmurty:
One problem I have noticed – and which I don't yet know how to solve – is how to make Git's `textconv` filter work with different password contexts.

Unlike all the other Git internals we rely on, `textconv` does not seem to have a way to access the original path of a file it is operating on because it is only provided with a temporary file copy. Without the original file path, we don't have a way to look up the context name for the file and therefore cannot find the right password. See the documentation [here](https://www.git-scm.com/docs/gitattributes#_performing_text_diffs_of_binary_files) where there is no mention of a `%f` option, like we have and rely on for `clean` `smudge` operations.

Up until now the contexts implementation has (accidentally) ignored this issue and always used the password of the default/unspecified context. This will be okay in some cases, but I can foresee nasty problems using `git diff` or `git log` for secret files in non-default contexts unless I can find a way to work around this limitation. The problem area of code is here: https://github.com/elasticdog/transcrypt/blob/multiple-passphrases-in-one-repo-wip/transcrypt#L213-L215

jmurty:
Hi @benclanet @acro5piano @yjarrar and any others interested in the contexts feature, can you please do some testing of the updated [multiple-passphrases-in-one-repo-wip](https://github.com/elasticdog/transcrypt/tree/multiple-passphrases-in-one-repo-wip) branch to confirm whether it works for you, and if you come across any issues?

I think that branch is in reasonable shape and it can be cleanly merged into `main` for longer-term testing, provided there aren't any major problems. However it's only lightly tested so far and includes non-trivial changes, so **be careful** not to test in vital repositories or in any way where bugs could lose valuable data.

Also note that I was forced to change the underlying approach due to the [`textconv` issue discussed above](https://github.com/elasticdog/transcrypt/pull/63#issuecomment-1279760401), so **the required _.gitattribute_ patterns are different**. Instead of using a single extra `crypt-context=some-name` attribute in the pattern, you must now use a pattern like the following that includes the context name in every attribute value:
```
# .gitattributes: Encrypt 'top-secret' file in the "super" context
top-secret filter=crypt-super diff=crypt-super merge=crypt-super
```

acro5piano:
Hi @jmurty Thank you so much for your effort. I tested the branch and confirmed it working correctly.

My test repository is here. I logged every commands for easy reproduction.
https://github.com/acro5piano/transcrypt-context-test

jmurty:
Thanks very much for testing this @acro5piano and for showing your work in the linked repo. And thanks for your patience too, I see a comment from you about this feature way back in 2020!

Your confirmation has given me the confidence to merge the new contexts feature to the `main` branch, see commit https://github.com/elasticdog/transcrypt/commit/e08c3595e6ec57bc448e50401cb7ed845d866419

We can keep testing it there and, all going well, roll it out in a 2.3.0 release in the not-too-distant future.

acro5piano:
@jmurty I'm glad to hear that. Thank you for your work!

Issue #68, opened 2019-08-28
Transcrypt version and help output not displayed outside git repositories


JensErat:
`transcrypt --version` and `transcrypt --help` should also work outside git repositories, as they're frequently used to test whether the script is installed correctly. They give no output, though.

Example terminal session:

```sh
➜  /tmp mkdir test
➜  /tmp cd test 
➜  test transcrypt --version
➜  test git init
Initialized empty Git repository in /tmp/test/.git/
➜  test git:(master) transcrypt --version
transcrypt 2.0.0
```

Same applies to `transcrypt --help`.

Expected behavior: version also displayed outside an initialized git repository.

apeschel:
I took a quick look, and this is happening due to `errexit` being set at the start of the script. My personal opinion is that errexit shouldn't be used at all, and that proper error handling should be relied on instead. It looks like the transcrypt has pretty good error handling throughout, so it may be fine to remove `errexit` entirely. Alternatively, turning on could be deferred until after `run_safety_checks`.

PR #70, opened 2019-11-12
Defer errexit to as late as possible

apeschel:
At the moment, transcrypt immediately exits with a failure when run
outside of a git repository, regardless of what command is given to
transcrypt. This causes unexpected behavior when running commands like
"help" or "version".

This commit moves errexit to as late as possible, while still keeping it
as a safe guard for the time being. Ideally, errexit would not be used
in transcrypt at all.

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euo pipefail
+shopt -so nounset pipefail
 
 #
 # transcrypt - https://github.com/elasticdog/transcrypt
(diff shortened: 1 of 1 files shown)


apeschel:
Fixes https://github.com/elasticdog/transcrypt/issues/68

apeschel:
@elasticdog Could you weigh in on this when possible, please?

elasticdog:
Hey @apeschel! Thank you for the PR, and my apologies on taking so long to respond. I think I've come up with a cleaner solution to issue #68 without having to shuffle around the errexit timing. There were a lot of changes to enable "strict" Bash within #53 that took me a while to warm up to, but now I can see the value (especially since transcrypt doesn't have a test suite we can rely on, it can help when adding new features).

Let me know if my other PR addresses the issues you were tackling here, and I'd also be curious to hear your thoughts on why you think errexit should be eliminated completely. Cheers.

elasticdog:
I'm going to close this one out, as the original issue was addressed via https://github.com/elasticdog/transcrypt/pull/71, but thanks again for taking the time to look at the problem...I'd be glad to discuss errexit with you further if you have strong feelings about it.

PR #71, opened 2019-12-08
Split repository metadata gathering into a function

### LINKED ISSUE #68: Transcrypt version and help output not displayed outside git repositories

elasticdog:
This will prevent the issue of error exiting for certain option flags when outside of a Git repository (like `--help` and `--version`). As part of this, we must also prevent error exits from certain Git commands so we can handle notifying the user more cleanly.

Resolves #68

cc: @JensErat @apeschel

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -21,28 +21,7 @@ readonly VERSION='2.0.0'
 # the default cipher to utilize
 readonly DEFAULT_CIPHER='aes-256-cbc'
 
-# regular expression used to test user input
-readonly YES_REGEX='^[Yy]$'
-
-## Repository Metadata
-
-# whether or not transcrypt is already configured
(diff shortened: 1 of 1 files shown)


apeschel:
@elasticdog  Much appreciated, thank you!

Issue #72, opened 2020-01-07
transcrypt does not update built-in clean/smudge filters


apeschel:
As far as I can tell, there's no code path that updates the transcrypt scripts stored in .git/crypt, and these files are only updated on initialization.

This is problematic, as changes to these scripts are never propagated out to repositories with transcrypt already initialized.

In https://github.com/elasticdog/transcrypt/commit/c8e4f7279972555282a74c96d7ce48fdbc930083 the method for generating the salt was modified, and is not backwards compatible with the previous method. As a result, users who initialized a repository with transcrypt previous to this commit will be generating different salts, and thus be producing different output than those who initialized the repository after this commit.

For now, the work around is to have all users run `transcrypt -u` and rerun `transcrypt`, but this is a very tedious exercise to orchestrate across any organization of any size, and is prone to error.

transcrypt should have some facility to update these files automatically and transparently to users.

The most straightforward approach I can think of to fix this in transcrypt is to have it update these files every time it is run in a git repository, but that might be overkill.

A better solution might be to have these files be replaced with either symlinks or simple wrappers that call back out to transcrypt, and add this functionality in as a sub-module to the transcrypt program. This way these functions would always be tied in to the correct version expected by the current version of transcrypt.

jmurty:
Hi @apeschel the new `--upgrade` command added in https://github.com/elasticdog/transcrypt/pull/86 might help with updates. It's still a manual step, but easier than the uninstall-then-reinit process.

jmurty:
Hi @apeschel I think the new `--upgrade` command added to the master branch in ea4aa35c solves the main problem outlined in this issue, so I am closing this specific issue. Please let us know what you think of it if you are able to try the `--upgrade` command now on the *master* branch.

Your idea about "internalising" the various clean / smudge / merge scripts as mentioned here and explored in detail in PR #73 is an interesting one, and something I'm still considering in a broader context where we may need more sophisticated and (hopefully) shared code to handle different versions and capabilities in OpenSSL for #55 etc.

Thanks for contributing!

PR #73, opened 2020-01-07
Internalize transcrypt crypt functions

apeschel:
At the moment, the crypt functions are initialized and permanently set
to whatever the version of transcrypt the user is using contains. This
means updates to these scripts are never propagated out to existing
users.

This commit moves the crypt functions into transcrypt itself, which
should allow for more transparent updates to the scripts to be
propagated out to users.

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -268,68 +268,55 @@ stage_rekeyed_files() {
 	fi
 }
 
-# save helper scripts under the repository's git directory
-save_helper_scripts() {
-	mkdir -p "${GIT_DIR}/crypt"
-
-	# The `decryption -> encryption` process on an unchanged file must be
-	# deterministic for everything to work transparently. To do that, the same
-	# salt must be used each time we encrypt the same file. An HMAC has been
-	# proven to be a PRF, so we generate an HMAC-SHA256 for each decrypted file
-	# (keyed with a combination of the filename and transcrypt password), and
-	# then use the last 16 bytes of that HMAC for the file's unique salt.
-
-	cat <<-'EOF' >"${GIT_DIR}/crypt/clean"
-		#!/usr/bin/env bash
-		filename=$1
-		# ignore empty files
-		if [[ -s $filename ]]; then
-		  # cache STDIN to test if it's already encrypted
-		  tempfile=$(mktemp 2>/dev/null || mktemp -t tmp)
-		  trap 'rm -f "$tempfile"' EXIT
-		  tee "$tempfile" &>/dev/null
-		  # the first bytes of an encrypted file are always "Salted" in Base64
-		  read -n 8 firstbytes <"$tempfile"
-		  if [[ $firstbytes == "U2FsdGVk" ]]; then
-		    cat "$tempfile"
-		  else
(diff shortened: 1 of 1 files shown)


apeschel:
This is meant as a POC as one possible solution to this problem:

Several caveats on this at the moment:

* This has only been quickly and minimally tested, but appears to be functional
* There still needs to be logic added to handle git config drift.

It's still possible for the git config expected by transcrypt to drift over time. There should be some functionality that updates the config to handle drift. Maybe some logic that checks git.transcrypt.version, compares it to the user's transcrypt version and updates the config if there's the user is using a newer version?

apeschel:
Related to https://github.com/elasticdog/transcrypt/issues/72

apeschel:
This is just one potential approach, another option is to check git.transcrypt.version on every run, and rerun a subset of save_configuration if the user's transcrypt version is newer.

apeschel:
It may also make sense to have something like a .transcrypt directory that is saved in the repository, which could save the helper scripts, but also allow them to be consistent between users.

elasticdog:
I'll have to give this one some thought...the original intention was that you wouldn't have to keep `transcrypt` installed after the repository is configured and the filters would still work. If you want to ensure a consistent version, you can always just check the script into your repo and share that version across users.

The lack of ability to update things automatically is a legitimate concern, but that also means we don't have to bake in knowledge of version checking and compatibility, which could be complex. Breaking changes have only happened once, and it was a major release (2.x) with very explicit instructions on how to handle it. Even if the filters updated transparently, you'd still have to coordinate transcrypt upgrades across users for the repo to be functional for everyone.

apeschel:
@elasticdog Having the scripts be part of the repository is also a good solution, but with their current home under `.git/crypt`, there's no way to actually include them in the repository. Just moving them to a top level crypt directory or something would be a good alternative.

jmurty:
Hi @apeschel I shamelessly stole your approach in this PR with mine in #112 which I have just merged. I didn't find this PR at the time to credit you properly, but ended up re-implementing much of what you proposed here.

A key difference of #112 to this one is that the entire `transcrypt` script as run at init time is copied into the Git `crypt/` directory and the clean/smudge/etc filters refer to that copied file from that point on, as opposed to referring to the system-installed `transcrypt` version.

I believe the #112 works against your main intention here, which was to make it easier to update transcrypt in repositories by just updating the system-wide transcrypt version.

On the bad side, you still need to manually `--upgrade` each repo to update transcrypt's behaviour or fix bugs.
On the good side, upgrading the system-wide transcrypt version won't mysteriously change behaviour for all init-ed repos, and uninstalling the system-wide transcrypt won't break everything.

jmurty:
The work already merged in #112 and further improvements pending in #114 will internalise what were many transcrypt-installed scripts into multiple references to the single, original script that is copied into the Git directory on installation.

apeschel:
@jmurty Looks like a big improvement, and it should fix the issue we were hitting with different users having different version of the crypt functions, even when they had the same version of transcrypt. Thanks!

PR #75, opened 2020-04-07
Add tests for core features using bats-core

jmurty:
This branch is now a candidate for merge


jmurty:
Hi @elasticdog would you mind setting up the Travis / GitHub integration as described here? https://docs.travis-ci.com/user/tutorial/#to-get-started-with-travis-ci-using-github

I can't do it because I don't have, and otherwise don't need, admin privileges for this repo.

Once this is set up I should be able to get automatic continuous integration tests working for this project.

tests/_test_helper.bash
```diff
@@ -0,0 +1,13 @@
+function init_git_repo {
+  git init $BATS_TEST_DIRNAME
+}
+
+function nuke_git_repo {
+  rm -fR $BATS_TEST_DIRNAME/.git
+}
```
elasticdog, line 7: Should this test for existence of a value in the `$BATS_TEST_DIRNAME` for safety before forcefully removing files recursively?

jmurty: Hi @elasticdog do you think the changes in 18b1aae6 are enough to ensure safety?

I tried to do something more clever there and fail tests where the safety checks failed, but I couldn't make it work in BATS-core so gave up and just logged a warning and do nothing if a safety check doesn't pass.

jmurty:
Woohoo, tests pass in Travis! https://travis-ci.com/github/elasticdog/transcrypt/builds/159603874

tests/test_not_inited.bats
```diff
@@ -0,0 +1,75 @@
...
+  skip "TODO: --list doesn't work for un-inited repo, should it?"
+  run ../transcrypt --list
+  [ "${lines[0]}" = "" ]
+}
+
+@test "not inited: no files listed for -l" {
+  skip "TODO: -l doesn't work for un-inited repo, should it?"
+  run ../transcrypt -l
+  [ "${lines[0]}" = "" ]
+}
```
jmurty, line 75: @elasticdog I'm not sure whether `transcrypt --list` should work in a repo that hasn't been initialised?
It did work in un-inited repos prior to 8ebb59ae, but didn't work in inited repos which was definitely a bug.

tests/_test_helper.bash
elasticdog: Yep, that's much better...thank you!

tests/test_not_inited.bats
elasticdog: Yes, it should work since it's just checking for the `crypt` attribute and doesn't rely on any other existing config. I believe as long as we add `requires_clean_repo=''`, then it will work in either scenario.

transcrypt
```diff
@@ -771,6 +771,8 @@ while [[ "${1:-}" != '' ]]; do
 		;;
 	-l | --list)
 		list='true'
+		requires_existing_config='true'
```
elasticdog, line 774:
```suggestion
```

jmurty: @elasticdog I couldn't get `--list` to work for both configured and unconfigured repos without adding a new flag `ignore_config_status` in ae5d9f61.

I might be missing something, but the flags and associated logic in `run_safety_checks()` made it impossible to express that an operation should work regardless of the config status of the repo.

Assuming the change in ae5d9f61 is okay, this branch can be merged now and future tests etc be added directly on the `master` branch.

jmurty:
@elasticdog Hmm, the Travis integration seems to be broken, I didn't get a test run for my latest commit at https://travis-ci.com/github/elasticdog/transcrypt/branches

Regardless, the tests run for me locally and I still think this PR is ready for merge as long as you're happy with my change in ae5d9f6

jmurty:
Travis got there in the end: https://travis-ci.com/github/elasticdog/transcrypt/builds/159920764

Issue #31, opened 2016-09-29
Avoid accidentally committing unencrypted files.


mrmachine:
I use the GitUp client on OS X, and I just discovered that it does not execute attribute filters (https://github.com/git-up/GitUp/issues/224).

If I'm not careful (or someone else using GitUp isn't careful), it is very easy to stage and commit an encrypted file from GitUp.

Could transcrypt automatically configure a pre/post-commit or pre-push hook to ensure that any files that should be encrypted, are always encrypted?

elasticdog:
I'm not quite sure how that would work. Even if it could automatically create a client-side hook, how could it test if the files are encrypted? The [clean filter script](https://github.com/elasticdog/transcrypt/blob/1c459bf0ccc25255c9706be751b913516138e1e1/transcrypt#L269-L271) does some naive testing for encrypted files, but there are cases where it doesn't hold up (like if an empty file is set to be encrypted). There could also be false-positives if you try and encrypt an already-encrypted file.

I'm also not familiar with the internal workings of GitUp, but it looks like [it runs hooks itself](https://github.com/git-up/GitUp/wiki) so there might be some odd interactions if the hook were to try and shell out to normal Git/transcrypt commands in order to check the attributes. Give me a bit to think about this...

mrmachine:
We switched away to git-secret to avoid accidentally committing unencrypted files, but it has it's own problems (GPG2 pain, no diffs in commit history, etc.)

So we're thinking about switching back, but we are still worried about how easy it is to accidentally commit unencrypted sensitive data with the wrong Git client.

Did you have any other ideas about how we might be able to minimise that risk?

It seems we can't do server side hooks on GitHub, so we'd have to tell everyone to manually setup a local hook, which is probably just as ineffective as telling people to be careful when using GitUp.

Would it be possible to have transcrypt have a less transparent mode of operation that stores the encrypted version in a `*.secret` file and adds the unencrypted file to `.gitignore` to ensure it is never accidentally committed?

It would be OK if we had to manually run commands (e.g. `transcrypt hide`, `transcrypt reveal`) to encrypted and decrypt the files.

elasticdog:
[...]

PR #77, opened 2020-04-14
Add Git pre-commit hook to abort commit with unencrypted files (#31)

jmurty:
On init, add a pre-commit Git hook script to check Transcrypt-managed files and abort a commit if there is an un-encrypted file staged in the index that would otherwise be committed in plaintext. See related issue #31

This is a safety mechanism to prevent accidental commits of plain text files that have been staged by tools that do not respect or run the *.gitattribute* filters that Transcrypt needs to do its job.

On commit failure, the error message says how to re-stage the file using Git on the command line:

    Transcrypt managed file is not encrypted in the Git index: secret_file

    You probably staged this file using a tool that does not apply .gitattribute filters as required by Transcrypt.

    Fix this by re-staging the file with a compatible tool or with Git on the command line:

        git reset -- secret_file
        git add secret_file

Because Git hooks work with single scripts only it is difficult to cleanly install and uninstall hook scripts, especially if the user already has a pre-commit hook script in place.

To handle this situation cleanly if naively

- always install a *pre-commit-crypt* copy of the hook script, which isn't active because it doesn't have the *pre-commit* file name
- on init, check if the user already has a *pre-commit* script and instead of clobbering it, print a warning telling the user to manually install the script from *pre-commit-crypt*
- on uninstall, only delete the *pre-commit* if it exactly matches *pre-commit-crypt* file, which would indicate the user had manually edited the *pre-commit* file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ complexity here would be worth it given transcrypt's use case.
 transcrypt is provided under the terms of the
 [MIT License](https://en.wikipedia.org/wiki/MIT_License).
 
-Copyright &copy; 2014-2019, [Aaron Bull Schaefer](mailto:aaron@elasticdog.com).
+Copyright &copy; 2014-2020, [Aaron Bull Schaefer](mailto:aaron@elasticdog.com).
 
 ## Contributing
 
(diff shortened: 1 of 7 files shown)


jmurty:
This branch is a mess of merges and spurious commits sorry, I should have cherry-picked instead of merging the original v1 changes.

If/when this is merged to master the whole lot should be squashed.

README.md
```diff
@@ -308,3 +308,14 @@ To run the tests:
...
+## Changes
+
+Improvements:
+
+- Add Git pre-commit hook to reject commit of file that should be encrypted but
+  has plain text content: a safety mechanism to prevent accidental commits of
+  plain text files staged by tools that do not respect the .gitattribute
+  filters Transcrypt needs to do its job.
+
+- Add functional tests.
```
elasticdog, line 321: Same suggestion as the other PR, but I think we could use a legit CHANGELOG.

jmurty:
Hi @elasticdog I have rebased and cleaned up the commits in this PR to make more sense, I think this is ready for review.

Regarding the change log I agree a CHANGELOG file would be best. This is probably something to tackle after dealing with this PR and #76 since the new changeling file should probably include all changes, not just changes specific to these PRs.

PR #79, opened 2020-04-27
(WIP) Experiment to run tests with GitHub Actions

jmurty:
Let's see if we can get tests to run on GitHub like with Travis.


.github/workflows/run-bats-core-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+        git clone https://github.com/bats-core/bats-core.git /tmp/bats-core-repo
+        mkdir -p /tmp/bats-core
+        bash /tmp/bats-core-repo/install.sh /tmp/bats-core
+
+    # Need to configure git globally to avoid errors when we create and use
+    # temporary test-specific Git repos fatal: `empty ident name`
+    - name: Configure dummy Git user.[name,email] values
+      run: |
+        git config --global user.name github-actions-user
+        git config --global user.email github-actions-user@nowhere.com
```
elasticdog, line 26: Minor point, but I'd suggest using one of the IANA-reserved ["Special-Use Domain Names"](https://tools.ietf.org/html/rfc6761) like `example.com` or `.invalid` for this purpose, so it doesn't end up getting tied to a legit domain that someone owns.

.github/workflows/run-bats-core-tests.yml
```diff
@@ -0,0 +1,29 @@
+name: Tests
+
+on: [push, pull_request]
```
elasticdog, line 3: Should we limit the push testing to just the master branch so PR testing isn't run twice?

.github/workflows/run-bats-core-tests.yml
jmurty: Done

jmurty:
@elasticdog 👍  Already [merged to master](https://github.com/elasticdog/transcrypt/commit/9fd84ff079a82aaca970a07d870561a2e1d68582) and done. I forgot to manually close this since I squash-merged away my many failed attempts

Issue #78, opened 2020-04-26
Windows issues, work-arounds and a minor change request


perost-l14:
Hi.
I am using this on Windows 10 and after some issues, it is working. For it to work for others, I think my change should be put back into the tool.

First the two hints -

- Set core.quotePath to false (git config --add core.quotePath false). Without this Danish (and similar) characters are escaped to \octal and that breaks further processing of those files
- Next, to install transact on Windows without bash, I simply added it as a git alias and use the built-in bash interpretator in git. This is done with git config --add alias.transcrypt = !~/path/to/transcrypt/transcrypt. After this, I can install it with 'git transcrypt'.

And then the required fix in transcrypt itself.

The input file seperator (IFS) must be set to linefeed to avoid having the bash for statement split the input into words on space. Any files I have with spaces were not processed. So I added IFS=$'\n' before the "for secret_file in" loops (just before https://github.com/elasticdog/transcrypt/blob/master/transcrypt#L338)

Hope this change can get in.

PS. A great tool. Looking at using it for client-side encrypted backup.

jmurty:
Thanks for the detailed issue report @perost-l14 

Could you please include here some example file names that caused you problems, especially those with spaces in the name and with non-ASCII Danish characters?

I suspect a couple of the problems you faced may not be due to running on Windows specifically, but flaws in the script when handling these kinds of file names. 

Also, any extra information you can provide about how exactly you are running `transcrypt` could be helpful.

perost-l14:
Thanks for the prompt response. Here is one of my files:
```
live/sub/2/Mig - røve.PNG
```
with
```
git config --add --local core.quotePath true
git ls-files live/sub/2/M*
"live/sub/2/Mig - r\303\270ve.PNG"
```
And the space in the file name messes up 'for secret_file in' as it splits it into words (the IFS thing).

I reran a test on whether core.quotePath has any impact at all. It does. Most things works; however, encryption fails -
```
*** WARNING : deprecated key derivation used.
Using -iter or -pbkdf2 would be better.
fatal: Unimplemented pathspec magic '"' in ':"live/sub/2/Mig - r\303\270ve.PNG"'
```
Note the -iter thing. There is another open issue on that.

My plan for this is -
- Create a repo on AWS CodeCommit (or somewhere else)
- Create 'staging' local .git outside my normal work folders
- Enabled transcrypt
- Add worktree to that to a work folder
- Adjust .gitignore and .gitattributes according to my requirements
[...]

PR #80, opened 2020-05-04
WIP Improve handling of challenging file names when 'core.quotePath=true'

jmurty:
As described in issue #78 transcrypt could fail to properly handle:

- file names with non-ASCII characters when Git's configuration
  option 'core.quotePath' is set to 'true' (as it is by default)
- file names containing spaces

This change improves handling of non-ASCII or space characters in
filenames and adds a test to exercise and confirm the fix.

Changes in particular:

- fix handling of space characters in file name in pre-commit hook
- override repository's 'core.quotePath' setting in commands that
  output paths to ensure 'core.quotePath=false' is always applied
  even when 'core.quotePath' is explicitly or implicitly (default)
  set to true.

### DIFF
--- a/tests/_test_helper.bash
+++ b/tests/_test_helper.bash
@@ -30,7 +30,7 @@ function nuke_git_repo {
 function cleanup_all {
   nuke_git_repo
   rm $BATS_TEST_DIRNAME/.gitattributes
-  rm $BATS_TEST_DIRNAME/sensitive_file
+  rm -f $BATS_TEST_DIRNAME/sensitive_file
 }
 
 function init_transcrypt {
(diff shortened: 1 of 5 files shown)


jmurty:
Thanks for the feedback @elasticdog. I too am unhappy with having `-c` options sprinkled through the code. I have been thinking through alternatives that still fix the underlying issue, but haven't come up with any great alternatives.

Options I have considered:

1. Forcibly set `core.quotePath=false` in transcrypt-enabled repositories. This would solve the problem, but in a somewhat aggressive way which might break peoples workflows or tooling. And it seems like a bad idea to override a Git default value like that unless it's really needed.

2. Replace the sprinkled `-c` overrides with Git aliases which versions that include the `-c core.quotePath=false` override, e.g: `ls-files-unquote`, `check-attr-unquote`, `ls-tree-unquote`. The `transcrypt` script would then use these alternatives.

   This is cleaner since it sets the overrides in one place, and works okay in my initial testing except it breaks usage of `transcrypt` in un-inited repos for actions like `transcrypt --list`.

PR #83, opened 2020-05-19
Improve merge driver script to work for complex merges

jmurty:
Improve the merge driver script to fix a problem that could occur
with complex merges where an encrypted file that couldn't be merged
cleanly would cause the entire merge to fail with the message:

    error: Your local changes to the following files would be overwritten by merge:

        sensitive_file

The bug was caused by being too aggressive in overwriting the working-
copy file $5 with merged content. Doing this early and in all cases
could leave what looks like manually modified files in the working
copy that Git would notice part-way through a multi-step merge and
abort.

The fix is to overwrite the working-copy $5 file only when a merge
is successful (no conflicts) at which point we really need that
specific $5 filename to contain the merged data to encrypt it.
In all other cases, we now leave the working-copy $5 file untouched
and only modify the temporary file $2 which is the file Git expects
to be managed by merge driver scripts.

Other minor improvements in this change are:
- quote ALL uses of $x file name/path arguments to guard against
  failures due to spaces in file paths
- simplify if/then logic to exit fast when needed
- try to better explain in comments what on earth is happening in
  the merge driver script and why.

### DIFF
--- a/tests/test_merge.bats
+++ b/tests/test_merge.bats
@@ -31,7 +31,8 @@ load $BATS_TEST_DIRNAME/_test_helper.bash
   git commit -m "Add line 2, change line 1"
 
   git checkout -
-  git merge branch-2
+  run git merge branch-2
+  [ "$status" -eq 0 ]
 
   run cat sensitive_file
   [ "$status" -eq 0 ]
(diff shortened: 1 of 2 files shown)


jmurty:
Hi @elasticdog I hit a problem with my merge driver script addition today in real-world usage, and this is the fix. I'll merge this straight to master unless you see any problems?

PR #87, opened 2020-05-21
Fix for newly-introduced bug in merge driver

jmurty:


jmurty:
This PR is already manually merged to master

Issue #74, opened 2020-02-21
Unencrypted data left behind after credentials are flushed


TheLocehiliosan:
Steps to reproduce:

1. Clone transcrypt-enabled repo
2. Run `transcrypt` with valid credentials
3. Run `git log -p` (displaying all change sets, included encrypted ones)
4. Run `transcrypt -f` (flushing the credentials)
5.
    A.  Run `for object in $(cat .git/refs/notes/textconv/crypt); do git show $object; done` (you'll see previously cached unencrypted data)

    or

    B.  Run `transcrypt` with **invalid credentials** (encrypted data should NOT be available at this time)
        Run `git log -p` (you'll see previously cached unencrypted data)

I looked into this, and the reason this data is left unencrypted is due to the setting `diff.crypt.cachetextconv=true`. I image this has been done for efficiency/speed, but the result is unencrypted data left behind. I can think of two ways to handle this problem.

1. Make `cachetextconv=true` optional, defaulting to `false`. This might be the simplest fix. Today this is hard-coded to `true`.
2. Upon `--flush`, hunt down any cached data. This data can be found by inspecting `<GIT_DIR>/refs/notes/textconv/crypt`. That will contain a list of git objects that need to be removed. This is probably more difficult, but might be best because then caching could be used.

jmurty:
Hi @TheLocehiliosan thanks for the report and reproduction steps. I confirmed the issue, and found that it also occurs when you uninstall transcrypt which also removes the creds and shouldn't leave behind (non-obvious) plaintext data.

I have applied a fix in the branch [74-clear-cached-plaintext-when-creds-flushed](https://github.com/elasticdog/transcrypt/tree/74-clear-cached-plaintext-when-creds-flushed), could you confirm that running `transcrypt -f` with the *transcrypt* script from that branch removes the unencrypted data?

The fix was relatively simple, though it does require a `git prune` to properly remove the plaintext objects which may not be safe to do without warning the user: 
https://github.com/elasticdog/transcrypt/pull/89/commits/80e9498f8cbc1e740a23847c915df7f7000eda3f#diff-2080fe2b0bfa1edd69c751a9d3978fb5R498-R501

TheLocehiliosan:
@jmurty - This is pretty close, but I think depending on how long the transcrypt credentials have been present, it is still possible for encrypted data to be present. `git prune` doesn't effect unreachable objects which have already been packed. I verified this by adjusting my reproduction steps like so:

[...]

PR #89, opened 2020-05-30
Remove cached plaintext of secret files on uninstall or flush

jmurty:
Fix issue #74 where flushing transcrypt credentials or uninstalling can leave cached plaintext versions of secret files in Git's object database with relatively easy-to-find refs.

### DIFF
--- a/tests/test_cleanup.bats
+++ b/tests/test_cleanup.bats
@@ -1,0 +1,92 @@
+#!/usr/bin/env bats
+
+load $BATS_TEST_DIRNAME/_test_helper.bash
+
+SECRET_CONTENT="My secret content"
+SECRET_CONTENT_ENC="U2FsdGVkX1/kkWK36bn3fbq5DY2d+JXL2YWoN/eoXA1XJZEk9JS7j/856rXK9gPn"
+
+@test "cleanup: transcrypt -f flush clears cached plaintext" {
+  encrypt_named_file sensitive_file "$SECRET_CONTENT"
+
+  # Confirm working copy file is decrypted
+  run cat sensitive_file
+  [ "$status" -eq 0 ]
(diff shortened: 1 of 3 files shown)


jmurty:
Hi @elasticdog the relatively simple fix in 80e9498 solves the problem, though I'm not sure how safe it is to run an explicit `git prune` without warning the user?

tests/test_cleanup.bats
```diff
@@ -0,0 +1,102 @@
...
+  [[ "${output}" = *"+$SECRET_CONTENT" ]]  # Check last line of patch
+
+  # Look up notes ref to cached plaintext
+  [[ -f $BATS_TEST_DIRNAME/.git/refs/notes/textconv/crypt ]]
+  cached_plaintext_obj=$(cat $BATS_TEST_DIRNAME/.git/refs/notes/textconv/crypt)
+
+  # Confirm plaintext is cached
+  run git show "$cached_plaintext_obj"
+  [ "$status" -eq 0 ]
+  [[ "${output}" = *"+$SECRET_CONTENT" ]]  # Check last line of patch
```
elasticdog, line 28: Is there significance to the inconsistent use of `[` versus `[[` for tests? I see it in some of the upstream bats-core examples as well, but wasn't sure if it's intentional.

jmurty: I have been following what seemed like common practice for bats and used single `[` `]` by default, and only "upgraded" to more powerful tests with double `[[` `]]` when it is necessary, as it is for the wildcard output matching in this example.

I can't think of a good reason to keep this distinction. I'll add a task to the prep-for-release PR to convert all single brackets in tests to double brackets for consistency, and to make it slightly less fiddly to write tests.

PR #90, opened 2020-06-02
Fix unit tests to pass for newer OpenSSL versions

jmurty:
Adjust a couple of output checks to look for expected text anywhere
in the output, instead of at specific lines, so tests will pass
despite warning output like this:

See also #55


jmurty:
Merged with a manual rebase

Issue #94, opened 2020-09-20
"tail -c 16" syntax in crypt/clean throws an error


ljm42:
```
# git add myfile
tail: cannot open '16' for reading: No such file or directory
hex string is too short, padding with zero bytes to length
*** WARNING : deprecated key derivation used.
Using -iter or -pbkdf2 would be better.

```
The issue is in this line of crypt/clean:
https://github.com/elasticdog/transcrypt/blob/0396245e6a1e238c4a08d0b89271f2490c4ec559/transcrypt#L299

On my system at least, "tail -c 16" is not valid, but other options work:

```
# echo "09876543210987654321" | tail -c 16
tail: cannot open '16' for reading: No such file or directory
# echo "09876543210987654321" | tail -c16
543210987654321
# echo "09876543210987654321" | tail --bytes=16
543210987654321

```
FYI:
```
# tail --version
tail (GNU coreutils) 8.32

```

jmurty:
Hi @ljm42 thanks for the report, that's a surprising edge case.

Are you new to using `transcrypt`? I don't think the format of the `tail -c` command is something we have changed recently, but maybe that's been a problem case for a while and we didn't realise.

And can you please check whether a command like `head -c 8` has the same problem on your system? We use that command in the pre-commit hook that's new in version 2.1.0.

jmurty:
Hi @ljm42 could you also please test the potential fix on this branch? https://github.com/elasticdog/transcrypt/compare/94-more-compatible-head-and-tail-bytes-commands

That branch removes the space between the `-c` switch and value for the `tail` command you mention, and for a similar `head` command in the pre-commit hook script.

ljm42:
Thanks! yes I'm new to transcrypt :)

Interestingly, `head` doesn't have the same issues as `tail`:
```
#echo "12345678901234567890" | head -c 8
12345678
#echo "12345678901234567890" | head -c8
12345678
#echo "12345678901234567890" | head --bytes=8
12345678

# head --version
head (GNU coreutils) 8.32
```

This version of the file:
  https://raw.githubusercontent.com/elasticdog/transcrypt/25c3c66522d4cc56e11a32adcb8dbcb9458bb7f9/transcrypt
does allow me to "git add" without the tail error, thanks!

```
# git add testfile
*** WARNING : deprecated key derivation used.
Using -iter or -pbkdf2 would be better.
```

But I'm not sure that everything is working yet. Should "git show" automatically run the smudge filter? It does not:

```
[...]

PR #95, opened 2020-09-20
Improve compatibility of head & tail commands with -c option #94

jmurty:
Remove the space between the `-c` argument and the number of
characters/bytes to output, because the space between the switch and
the value is incompatible with tail in GNU coreutils 8.32


elasticdog:
Thanks @jmurty!

PR #97, opened 2020-09-24
add test for git show --textconv

ljm42:


jmurty:
Thanks for this @ljm42

PR #98, opened 2020-09-24
prevent tests from failing if global git name/email aren't set

ljm42:


jmurty:
Thanks very much @ljm42, with this I should be able to get rid of this workaround: https://github.com/elasticdog/transcrypt/blob/0ddc0e24ad68186598ac247d794eb37fc1d327d3/.github/workflows/run-bats-core-tests.yml#L32-L38

PR #99, opened 2020-09-25
skip symlinks in pre-commit-crypt hook

dgo-:
* git show :$FILE contains for symlinks the location of the target,
  because of this they are show as not encrypted.


jmurty:
Thanks very much for this fix @dgo- I added a unit test to confirm the fix and merged it in https://github.com/elasticdog/transcrypt/commit/0ddc0e24ad68186598ac247d794eb37fc1d327d3

PR #107, opened 2020-12-28
Fix error in zsh completion

erijo:
{} should only be used when there are multiple options. Otherwise you
get the following error:

  _arguments:comparguments:325: invalid argument: (--upgrade -c
  --cipher -d --display -f --flush-credentials -p --password -r
  --rekey){--upgrade}[upgrade transcrypt]


jmurty:
Thanks for this fix @erijo, it will be included in the next release

Issue #37, opened 2017-03-10
Detection of clean repo is not reliable.


mrmachine:
I'm using codefresh to build Docker images (including the `.git` directory), with an entrypoint script that configures the git repo for transcrypt with a password from the environment, to decrypt secret files at runtime.

transcrypt thinks the git repo is dirty:

```bash
# transcrypt -c "${TRANSCRYPT_CIPHER:-aes-256-cbc}" -p "$TRANSCRYPT_PASSWORD" -y
transcrypt: the repo is dirty; commit or stash your changes before running transcrypt
```

When I run the `git diff-index` command that transcrypt uses internally (minus the `--quiet` arg), it seems to return some odd results:

```bash
# git diff-index HEAD -- | tail                                                                                                   
:100644 100644 3d6e0e93c2bfef67162836b2f0b5ad24c526929e 0000000000000000000000000000000000000000 M      staticroot/apple-touch-icon-precomposed.png        
:100644 100644 9089ef7ffea1efe2de79b49b5e6a5856360de4d6 0000000000000000000000000000000000000000 M      staticroot/apple-touch-icon.png                    
:100644 100644 7b74f8caf013ba164ca9049f7bd4eea0555ff720 0000000000000000000000000000000000000000 M      staticroot/favicon.ico                             
:100644 100644 5d00a90141a920a278e2a8239e560b7ea395a1db 0000000000000000000000000000000000000000 M      staticroot/humans.txt                              
:100644 100644 214e4119653f9c6a4c48cd0ebb06a6754f00f62b 0000000000000000000000000000000000000000 M      staticroot/robots.txt                              
```

But the repo *is* clean, cloned and built by codefresh:

```bash
# git status                                                                                                                      
HEAD detached at 153c57c                                                                                                                                   
nothing to commit, working directory clean                                                                                                                 
```

Strangely, after running `git status`, transcrypt suddenly agrees that the repo is clean and agrees to configure the repo:

```bash
# git diff-index HEAD -- | tail                                                                                                   

# transcrypt -c "${TRANSCRYPT_CIPHER:-aes-256-cbc}" -p "$TRANSCRYPT_PASSWORD" -y
The repository has been successfully configured by transcrypt.                                                                                             
```

Is there something wrong with the way codefresh is cloning the repo before it builds the Docker image?

Is there a more reliable way transcrypt can detect a dirty repo (as reported by `git status`)?

elasticdog:
[...]

PR #109, opened 2021-02-03
Ensure Git index is up-to-date before dirty repo  check #37

jmurty:
Should fix/avoid failures seen in CI systems where the repo seems
dirty when it really isn't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ The format is based on [Keep a Changelog][1], and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- Ensure Git index is up-to-date before checking for dirty repo, to avoid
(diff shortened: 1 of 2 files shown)


elasticdog (approved):
Nice find on the `--really-refresh` flag...this LGTM, assuming that someone from #37 can test and confirm.

elasticdog:
This got me thinking...I wonder if this technique could replace some of the times when we "touch all encrypted files to prevent stale stat info"?

andreineculau:
> This got me thinking...I wonder if this technique could replace some of the times when we "touch all encrypted files to prevent stale stat info"?

👏 I thought the exact same thing! I never got to test it though, but based on how I understand this to work, it should do the job just fine.

jmurty:
Getting this moving, it should be a pretty safe change even though we haven't receive direct feedback that it fixes the problem

Issue #104, opened 2020-12-01
transcrypt ignores the `git config core.hooksPath` setting


ljm42:
By default, git runs hooks from the `.git/hooks` directory, but it allows for the directory to be adjusted through the `core.hooksPath` setting, like this:
  `git config core.hooksPath "/some/other/path"`

The PR below updates transcrypt to use the `core.hooksPath` setting if it exists, otherwise it defaults to `.git/hooks`

This change should be transparent, except that I did add the GIT_HOOKS directory to the `--display` output. If this is unwanted, the additional output could be suppressed when it contains the default value.

In my case I need to go a step further and change the location of the `crypt/[clean|smudge|textconv|merge]` files as well. The reason is that my git repo is on a drive that is mounted without execute perms, so git can't execute files stored in `.git/[hooks|crypt]`

If you are open to it, I will submit another PR that checks whether `core.hooksPath` is set. If so, it would create the `crypt` folder in `core.hooksPath`. Otherwise, keep it where it is (`.git/crypt`).  Does that sound reasonable?

jmurty:
Hi @ljm42 thanks for your submission. I have created a new pull request #110 which respects the `core.hooksPath` setting, without reporting on it under `--display`. I feel it's probably excessive to report on the `GIT_HOOKS` path even if it has a custom setting, since it isn't really specific to Transcript and the user is probably already aware of that customisation.

Regarding the custom location for the `crypt/` directory, I think that storing it under a hooks directory would be surprising and confusing to people even if this would only happen for those with custom `core.hooksPath` settings. Could this be done differently, perhaps with a new `--crypt-path` script option?

ljm42:
Thanks @jmurty, this looks great! A command line param for --crypt-path sounds like a good solution too

ljm42:
I like the changes you are working on over in https://github.com/elasticdog/transcrypt/pull/112

Would you be open to defining something like this:
`CRYPT_PATH="${GIT_DIR}/crypt"`
and then using that variable everywhere? It would make it much easier for me to tweak my local install, even without having the `--crypt-path` option

jmurty:
Hi @ljm42 thanks for the reminder, this is on my radar. I'm doing some further clean-up work based on the now-merge #112 PR which should make it easier still to customise the `crypt/` directory path.

jmurty:
[...]

PR #110, opened 2021-02-03
Respect git `core.hooksPath` setting when installing the pre-commit hook #104

jmurty:


elasticdog (approved):
Thanks @jmurty, this LGTM and I agree with splitting it out from the other changes in #104.

Issue #108, opened 2021-02-03
Suggestion: provide a mechanism for users to specify the openssl path


jwatt:
On macOS it would be useful to be able to tell `transcrypt` to use the Homebrew installed `openssl`, but I don't want to add `/usr/local/opt/openssl/bin` to `PATH` due to the potential for side affects. More specifically I'm hoping that `yadm` could pass any value set for `yadm.openssl-program` in its configuration file along to `transcrypt` in order to avoid the "deprecated key derivation used" warnings.

jmurty:
Hi @jwatt this is something we have discussed before in and around this comment: https://github.com/elasticdog/transcrypt/issues/55#issuecomment-637094456
So far we have decided against making the OpenSSL binary path configurable, though like you I would enjoy being able to customise the OpenSSL version used on MacOS.

I'm not familiar with `yadm` (Yet Another Dotfiles Manager). This seems to be the relevant section on using [`yadm` with transcrypt](https://yadm.io/docs/encryption#) though the only mention of the `yadm.openssl-program` option I found is [here](https://github.com/TheLocehiliosan/yadm/blob/master/yadm.md).

Can you talk more about how `yadm` could, or should, relay settings like `yadm.openssl-program` through to programs it invokes? If it does something relatively simple, we might be able to work from that to come up with a clean way to set a custom OpenSSL path both with and without using `yadm`.

jwatt:
Hi @jmurty. Thanks for the reply and link to the previous discussion.

Note that I'm just a casual user of yadm. That said, it seems like yadm is essentially a git wrapper that at its core invokes `git --git-dir=~/.local/share/yadm/repo.git --work-tree=~ <other args passed to yadm>` (plus other niceties).

yadm, like transcrypt, is just a single file shell script. I had expected that `yadm` invokes `transcrypt` frequently, passing command line arguments, and therefore the obvious mechanism for yadm to communicate the openssl path with transcrypt would be for it to pass the openssl path using a new `transcrypt` command line argument. It looks like I was mistaken in my assumptions about how things work though, and actually once transcrypt is set up yadm plays no direct part in the encryption/decryption or invoking transcrypt code.

Besides that, on reflection it probably wouldn't make sense to tell yadm about the openssl path if it's just passing on the args to transcrypt.

[...]

PR #111, opened 2021-02-12
WIP Let user set a custom path to openssl

jmurty:
This is work-in-progress. See  #108

Add the optional --openssl-path argument when initialising a repo to
tell transcrypt to use an explicit path to the openssl binary,
instead of using whatever version is on the user's path.

The openssl path is now saved as a new transcrypt.openssl-path Git
config local setting in the repository, alongside the other transcrypt
settings like cipher, password etc.

If the user provides --openssl-path this explicit path is stored in
the local Git config, otherwise the default value 'openssl' is stored
which will fall back to the default behaviour of finding openssl on
the user's $PATH.

The --openssl-path argument gets special treatment for upgrades: an
--openssl-path argument value given along with --upgrade will replace
the existing config setting, despite config settings normally being
retained across upgrades.

### DIFF
--- a/man/transcrypt.1.ronn
+++ b/man/transcrypt.1.ronn
@@ -25,6 +25,11 @@ The transcrypt source code and full documentation may be downloaded from
     the password to derive the key from;
     defaults to 30 random base64 characters
 
+  * `--openssl-path`=<path_to_openssl>:
+    use OpenSSL at this path; defaults to 'openssl' in $PATH.
+    To change after initialization:
+        git config transcrypt.openssl-path NEW_PATH
+
   * `-y`, `--yes`:
     assume yes and accept defaults for non-specified options
 
--- a/man/transcrypt.1.ronn
+++ b/man/transcrypt.1.ronn
@@ -25,10 +25,8 @@ The transcrypt source code and full documentation may be downloaded from
     the password to derive the key from;
     defaults to 30 random base64 characters
 
-  * `--openssl-path`=<path_to_openssl>:
-    use OpenSSL at this path; defaults to 'openssl' in $PATH.
-    To change after initialization:
-        git config transcrypt.openssl-path NEW_PATH
+  * `--set-openssl-path`=<path_to_openssl>:
+    use OpenSSL at this path; defaults to 'openssl' in $PATH
 
   * `-y`, `--yes`:
     assume yes and accept defaults for non-specified options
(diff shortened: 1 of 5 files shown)


transcrypt
```diff
@@ -973,6 +985,8 @@ rekey=''
 show_file=''
 uninstall=''
 upgrade=''
+openssl_path='openssl'
+openssl_path_given=''
```
elasticdog, line 989: You could probably just use a single variable and check if it's empty before setting the default?

transcrypt
```diff
@@ -912,6 +920,10 @@ help() {
 		     -i, --import-gpg=FILE
 		            import the password and cipher from a gpg encrypted file
 
+		     --openssl-path=PATH_TO_OPENSSL
+		            use a specific openssl binary instead of the version in \$PATH.
+		            Use this with --upgrade to update transcrypt's openssl path
+
```
elasticdog, line 926: The README and man page will also need a refresh with the new option flag information.

elasticdog (commented):
This is looking nice! I like the solution of storing the path in the config and can appreciate the troubles that some homebrew setups can introduce. I left a couple of comments.

transcrypt
jmurty: I reconsidered and dropped support for changing just the openssl path when `--upgrade`-ing a repo, since it was weird to make that setting the one and only user-defined setting that could be changed when upgrading. Instead, the help text now says how to change it using `git config`

transcrypt
jmurty: I have updated the README, changelog, and man pages. The help text itself could probably use some editing, it became log with the inclusion of the `git config` instruction

man/transcrypt.1.ronn
```diff
@@ -25,6 +25,11 @@ The transcrypt source code and full documentation may be downloaded from
     the password to derive the key from;
     defaults to 30 random base64 characters
 
+  * `--openssl-path`=<path_to_openssl>:
+    use OpenSSL at this path; defaults to 'openssl' in $PATH.
+    To change after initialization:
+        git config transcrypt.openssl-path NEW_PATH
```
elasticdog, line 31: Hmmm...it would be nice to not ever need to drop to raw `git config` commands at all. Would it make more sense to have this be named `--set-openssl-path` and allow it to be called whenever (during initialization or otherwise) to update the setting?

jmurty: That's a really good idea. I have made this change, and it makes it much clearer when `--set-openssl-path` can be used.

elasticdog (approved):
This looks great, thanks for iterating on it!

PR #112, opened 2021-02-19
WIP: Install transcrypt script into repo

jmurty:
When initializing transcrypt, copy the whole script file into the
repository and make it runnable with the `git transcrypt` alias.

This approach gives a number of benefits:

- You can always run transcrypt within an initialized repo, even if
  the wider system no longer has it installed
- When you run `git transcrypt` within a repo you are guaranteed to
  be running the appropriate version of the script, even if the
  system's version has changed
- The clean, smudge, textconv, and merge scripts become trivial,
  just invocations of the main script
- The old contents of the clean, smudge, textconv, and merge scripts
  become part of the main script, and thus check-able and lint-able

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -362,7 +362,8 @@ stage_rekeyed_files() {
 save_helper_scripts() {
 	mkdir -p "${GIT_DIR}/crypt"
 
-	local current_transcrypt=$(realpath "$0" 2>/dev/null)
+	local current_transcrypt
+	current_transcrypt=$(realpath "$0" 2>/dev/null)
 	cp "$current_transcrypt" "${GIT_DIR}/crypt/transcrypt"
 
 	cat <<-'EOF' >"${GIT_DIR}/crypt/clean"
--- a/transcrypt
+++ b/transcrypt
@@ -108,6 +108,91 @@ die() {
 	exit "$st"
 }
 
+# The `decryption -> encryption` process on an unchanged file must be
+# deterministic for everything to work transparently. To do that, the same
+# salt must be used each time we encrypt the same file. An HMAC has been
+# proven to be a PRF, so we generate an HMAC-SHA256 for each decrypted file
+# (keyed with a combination of the filename and transcrypt password), and
+# then use the last 16 bytes of that HMAC for the file's unique salt.
+
+git_clean() {
+	filename=$1
+	# ignore empty files
+	if [[ ! -s $filename ]]; then
+		return
+	fi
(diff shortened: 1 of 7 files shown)


jmurty:
Hi @elasticdog what do you think about this approach of copying the entire `transcrypt` file into a repo on initialisation?

I was inspired by an idea from someone else's suggestion in an issue or PR from ages ago, I can't find the reference right now to give credit. But I think it's a good idea.

transcrypt
```diff
@@ -276,103 +362,31 @@ stage_rekeyed_files() {
 save_helper_scripts() {
 	mkdir -p "${GIT_DIR}/crypt"
 
-	# The `decryption -> encryption` process on an unchanged file must be
-	# deterministic for everything to work transparently. To do that, the same
-	# salt must be used each time we encrypt the same file. An HMAC has been
-	# proven to be a PRF, so we generate an HMAC-SHA256 for each decrypted file
-	# (keyed with a combination of the filename and transcrypt password), and
-	# then use the last 16 bytes of that HMAC for the file's unique salt.
+	local current_transcrypt="$0"
```
elasticdog, line 365: I think we'd want to use our `realpath` function here to ensure the copy will work even if `transcrypt` is elsewhere in the user's `$PATH`?

elasticdog (commented):
I like this idea and especially the advantages of making the filter scripts check-able and lint-able, but I do worry a bit about the user experience and potential confusion with multiple copies of the script being in use (when to use `transcrypt` versus `git transcrypt`), handling upgrades, etc..

Since this puts the script into the `.git/crypt/` directory and isn't tracked, there can still easily be version mismatches between clones of the repo. I almost wonder if we did something like this, but created a hidden (but tracked) `/.transcrypt` version of the script instead, if that would be an improvement?

I'm still thinking through the idea a bit, but those are my initial thoughts.

jmurty:
Could we avoid user confusion by copying the script into the crypt dir to use internally, without creating the Git alias or telling the user?

jmurty:
I have removed the `git transcrypt` alias, so the fact the original `transcrypt` script is copied into the repo's `crypt/` directory is now hidden from the user: and internal implementation detail they don't need to know or care about.

elasticdog (approved):
:heart_eyes: I like that this gets us some development advantages and remains transparent to the user. I'm still trying to mentally wrap my head around potential benefits of having a `git transcrypt` alias, but we can handle that independently from this change.

PR #114, opened 2021-02-27
Further clean-up of helper scripts

jmurty:
Make further improvements and clean up helper scripts, following the consolidation of these scripts into the single *transcrypt* owner script as of 561d158bb79fff95b1c37b195266990c97e6a6c3

### DIFF
--- a/.github/workflows/run-bats-core-tests.yml
+++ b/.github/workflows/run-bats-core-tests.yml
@@ -44,8 +44,8 @@ jobs:
       run: git version
 
     # Configure default Git branch name to suppress hint warnings
-    - name: Configure default Git branch to "master"
-      run: git config --global init.defaultBranch master
+    - name: Configure default Git branch to "main"
+      run: git config --global init.defaultBranch main
 
     - name: Install and set up bats-core
       run: |
--- a/.github/workflows/run-bats-core-tests.yml
+++ b/.github/workflows/run-bats-core-tests.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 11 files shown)


.github/workflows/run-bats-core-tests.yml
```diff
@@ -40,6 +40,13 @@ jobs:
     - name: Print OpenSSL version
       run: openssl version
 
+    - name: Print Git version
+      run: git version
+
+    # Configure default Git branch name to suppress hint warnings
+    - name: Configure default Git branch to "main"
+      run: git config --global init.defaultBranch main
```
elasticdog, line 48: I'd be glad to rename our default branch to `main` for the transcrypt repository itself as well...I'm not currently aware of anyone writing packages based on pulling `master`, so it shouldn't be terribly disruptive (and GitHub will let users know how to update their local clones).

elasticdog: We would have to update our GitHub Actions workflow to match the new branch name, but I think that would be it?

jmurty: Yes, let's do it! I think I have covered the necessary changes in https://github.com/elasticdog/transcrypt/pull/122

elasticdog: I've merged in #122 and updated the branch, but the tests will fail until we get this `defaultBranch` setting in place...I should have thought a bit more about the order of operations here, but not a huge deal.

elasticdog (approved):
I think you'll have to rebase to get the tests passing again (and deal with the merge conflicts), but otherwise this is looking good to me. I'm excited that all of the recent refactoring is making the project more testable and resilient! Thanks @jmurty!

jmurty:
This work is now squashed and merged to `main` in fdf81c53f0ad27651e03a67ea732b164d209e948

Issue #116, opened 2021-02-27
'clean' filter fails on small files


ljm42:
If you try to encrypt a small file (less than 8 chars) it fails:

```
# git add <filename>
fatal: <filename>: clean filter 'crypt' failed
```

The problem is with the `read` command in git_clean():
  https://github.com/elasticdog/transcrypt/blob/dce1ad0dc41d3fb5eb754e5e135761910f275968/transcrypt#L129

One solution is to add an `or` statement like this:
```
  read -rn 8 firstbytes <"$tempfile" || firstbytes=X
```

jmurty:
Hi @ljm42 can you please test the change in PR #117 to see if it fixes the problem with small files?

ljm42:
This is great! I really appreciate you looking into this, and for your work on #114  as well!

My dataset includes txz files, which are essentially zips that have been renamed. And apparently they contain null bytes in the first 8 chars, because they cause this line:
`firstbytes=$(head -c8 "$tempfile")`
to throw this error:
`transcrypt: line 134: warning: command substitution: ignored null byte in input`
(ignore the line number, I added some debug code to figure out which files were causing the error)

I solved it by converting null bytes to spaces:
`firstbytes=$(head -c8 "$tempfile" | tr '\0' ' ')`
and now `git_clean()` works on both small files and zip files.

It looks like all of the encryption unit tests are based on the string `My secret content`. Would it make sense to include basic tests for other variations such such as a string that is less than 8 characters, and a super small zip file? Just to ensure that these types of files will work in future versions?

jmurty:
Hi @ljm42 I have updated the PR #117 with a couple of extra tests for small files and files with a null byte near the start.

Can you check that the `fix-clean-filter-for-small-files` branch works for your problem files? If so, I should be able to merge what is a fairly minor change.

ljm42:
Sorry, #117 `fix-clean-filter-for-small-files` branch still throws warnings with txz files (although I think it does actually encrypt them). Here is a test with a small txz file (rather than the string from the new test `sh\0x`):

```
# echo "sample text" > secret.txt
# tar cfJ archive.txz secret.txt
# rm secret.txt
# echo "archive.txz filter=crypt diff=crypt merge=crypt" >> .gitattributes
# git add .gitattributes archive.txz
.git/crypt/transcrypt: line 129: warning: command substitution: ignored null byte in input
```
Running line 129 directly results in the same:

```
[...]

PR #117, opened 2021-03-01
Potential fix for handling small files

jmurty:
See #116 

The `read -rn 8` command doesn't seem to cope with small files, and
does all kinds of strange things when I try running it manually on
latest macOS to test, including:
- not reading anything at all from files with short first lines
- reading too much from files with long first lines (more than 8 chars)

Overall the `head -c8` approach used everywhere else to do this work
is probably safer, and should work with small files as well.


elasticdog (approved):
Thanks @jmurty for the fix here and for adding tests. Hopefully the user can confirm the fix, but this approach is good with me :+1:

Issue #118, opened 2021-03-31
Partial commits of encrypted files use the wrong salt


Aramgutang:
While `git add -p` doesn't work with transcrypt, and some GUIs (e.g. git-cola) will not let you stage hunks/lines of changes for encrypted files, such partial commits do appear to be possible from VS Code (and possibly other interfaces).

However, such commits have the unfortunate side effect of leaving the repo in a state where (after any uncommited changes are removed, or the commit is pulled on a different machine) `git status` reports the encrypted file as modified, however `git diff` shows no changes, and `git checkout filename` does nothing. This is quite frustrating, as it can prevent pulling, checking out other branches, etc, and can't be worked around with `git stash`.

The reason this happens is that an incorrectly constructed salt is used when partial changes are committed. Thus, when git puts the local decrypted file through `clean`, the resulting encrypted binary differs to the encrypted binary that has been committed, even though both files decrypt to produce the exact same output. So the encrypted files differ, but the decrypted files do not, resulting in git's confusion as described above.

The cause of this lies in the parameters passed to OpenSSL in `clean` to generate the salt. The digest uses the SHA256 hash of the file read from `$filename`, assuming its contents are the same as the input it received (and stored in `$tempfile`). But since the former file also contains changes that were not staged for the commit, the contents differ. Thus, when the commit is pulled on another machine, the salt will be generated from the hash of the file without those local changes, and will not match.

As far as I can tell, the fix for this is simple: just replace `$filename` with `$tempfile` when generating the salt. I'll make a pull request with the change and leave it to you to verify that it doesn't break anything else.

jmurty:
Thanks for the detailed report and debugging @Aramgutang 

If you create a pull request with your proposed fix the automated unit tests should run, giving us a quick check for any unexpected problems.

jmurty:
This fix is now merged via #119

PR #119, opened 2021-03-31
Fix incorrect salt when partially staged files are commited

Aramgutang:
See ticket #118.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog][1], and this project adheres to
 
 - Respect Git `core.hooksPath` setting when installing the pre-commit hook. (#104)
 - Zsh completion. (#107)
+- Fix salt generation for partial (patch) commits (#118)
 
 ## [2.1.0] - 2020-09-07
 
--- a/transcrypt
+++ b/transcrypt
@@ -133,7 +133,7 @@ git_clean() {
 		cipher=$(git config --get --local transcrypt.cipher)
 		password=$(git config --get --local transcrypt.password)
 		openssl_path=$(git config --get --local transcrypt.openssl-path)
-		salt=$("${openssl_path}" dgst -hmac "${filename}:${password}" -sha256 "$filename" | tr -d '\r\n' | tail -c16)
+		salt=$("${openssl_path}" dgst -hmac "${filename}:${password}" -sha256 "$tempfile" | tr -d '\r\n' | tail -c16)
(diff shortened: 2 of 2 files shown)


jmurty:
@Aramgutang Can you think of any way we could add unit test coverage for partial commits of encrypted files?

Since `git-p` refuses to work with transcrypt-encrypted files it might be tough to simulate the behaviour.

Aramgutang:
Yeah, I read through the `git add/apply/am` documentation, and couldn't find anything useable. Basically the scenario to reproduce requires invoking the `clean` filter for a file, while feeding it input that isn't actually the contents of the file, all within the context of generating a blob object for the change.

The various GUIs probably have their own implementations of this, which is why some seem to be able to handle staging partial changes on encrypted files, and some aren't.

Aramgutang:
I read through the code for git-cola, and it appears that the way it stages hunks/lines is by generating patch files, writing them to a tmpfile, and passing them to `git apply`:

https://github.com/git-cola/git-cola/blob/b65aced9c089183eb008cd8c276b3c1941ee1f87/cola/diffparse.py#L261
https://github.com/git-cola/git-cola/blob/ac7e533541b46f7de5e778132be11227639ba0a7/cola/cmds.py#L287

But then again, staging partial changes to encrypted files in git-cola doesn't actually work, and it's probably because `git apply` doesn't pass the content through the filters, and neither does git-cola before writing the patchfile (which was the code I was hoping to find).

I reckon the only viable way to approach it would be along the lines of:
- Commit encrypted file.txt with content "AA"
- Save its encrypted content for reference later
- Modify file.txt to contain "BB"
- Call the `clean` filter script directly, passing "BA" into its stdin, and "file.txt" as the filename argument
- Generate a patch file using the received output and the encrypted content of the committed version we saved from before
- `git apply` the patch to the index, but not the working tree (using `--cached`)
- Modify file.txt to contain "BA"
- Verify that `git status` lists file.txt as having unstaged changes (prefix `MM ` when run with `-s`), but `git diff --cached` outputs nothing
- With the fix in place, `git status` should not list file.txt as having unstaged changes (prefix `M `), and the diff should still be empty

Seems like a lot of work...

jmurty:
Whew, yeah, definitely a lot of work. For this one, knowing that the change hasn’t broken anything already tested (most features) is probably good enough. 

Thanks for the explanation @Aramgutang

jmurty:
Thanks for this fix @Aramgutang! 

I tested it manually with VSCode pre- and post-fix and found the behaviour exactly as you described.

It's weird that VSCode allows partial commits when other tools refuse to, but that's a pretty important Git "client" to work with.

PR #121, opened 2021-04-25
Fix grep: unrecognized option: line-regexp

dimitrov-adrian:
In busybox there is no line-regexp for grep, using -x as it is used in busybox, macos and gnugrep. Adding also -F to force plain text and non pattern search.


jmurty:
This looks good to me, thanks for the fix!

PR #135, opened 2022-05-14
Support OpenSSL 3 which has changed behaviour for "Salted__" prefix

### LINKED ISSUE #133: OpenSSL 3 will break transcrypt

jmurty:
Work in progress to fix #133

### DIFF
--- a/.github/workflows/run-bats-core-tests.yml
+++ b/.github/workflows/run-bats-core-tests.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, ubuntu-20.04, macos-latest]
+        os: [ubuntu-18.04, ubuntu-20.04, ubuntu-22.04, macos-latest]
 
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ The format is based on [Keep a Changelog][1], and this project adheres to
 
 ### Fixed
 
+- Remain compatible with OpenSSL versions 3 and above which changes the way
+  explicit salt values are expressed in ciphertext (#133)
 - Ensure Git index is up-to-date before checking for dirty repo, to avoid
   failures seen in CI systems where the repo seems dirty when it isn't. (#37)
 - Respect Git `core.hooksPath` setting when installing the pre-commit hook. (#104)
--- a/tests/test_crypt.bats
+++ b/tests/test_crypt.bats
@@ -62,7 +62,7 @@ function check_repo_is_clean {
 
   run ../transcrypt --list
   [ "$status" -eq 0 ]
-  [ "${lines[0]}" = "sensitive_file" ]
+  [[ "${output}" = *"sensitive_file" ]]
 }
 
 @test "crypt: transcrypt --uninstall leaves decrypted file and repo dirty" {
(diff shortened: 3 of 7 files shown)


jmurty:
Hi @Erotemic @elasticdog this fix for OpenSSL 3 compatibility is working, provably so in the `ubuntu-22.04` test workflow which runs against OpenSSL version 3.0.2.

The fix required yet another temporary file to hold the cipher text at encryption time, prior to base64 encoding so we can check for the presence of the `Salted__` prefix and add it when it's missing. Because bash variables can't store all possible bytes – no null bytes – I can't think of a better way to handle this.

Issue #134, opened 2022-05-08
Formalize specs for transcrypt configurations with pbkdf2 & configured salt. 


Erotemic:
This issue is to discuss how pbkdf2 and new methods for generating salt should be handled in #132.

### PBKDF2

At a high level the problem is that transcrypt could be far more secure by using pbkdf2. Right now if you have at least a 5-word diceware-style password you are probably ok, and with 6 words you should be ok, should be ok isn't enough. Regardless enabling pbkdf2 drastically increases security against brute force passwords, so we should be able to leverage that if we want. The tradeoff is that encrypt/decrypt round trips are slightly slower (although there are mitigations for this as mentioned in #132), and in practice it's not too bad.

Currently #132 allows the user to enable pbkdf2 by either

(1) they add the cli args`-pbkdf2` or `--use-pbkdf2=1`. I like the first arg as it mirrors openssl, but I'm not sure about the second arg. I think it is important to have a flag it is a key/value pair. (key/value pair CLIs are far easier to work with problematically). 

(2) if they do not use a CLI arg, then transcript prompts them, similar to what it does for cipher, and makes the default choice "0", or off.

Note that existing repos will have to be decrypted with the current default settings and then rekeyed to change any config variable. The value for pbkdf2 is set in a similar way when you rekey. 

Using --pbkdf2 is a no brainier. Currently openssl warns if you don't do it. The details about the exact names of the CLI args are minor details. But there is one detail that is important to note: the current salt method that transcrypt uses renders the security of pbkdf2 moot. This brings us to the next topic, the salt method and ultimately the new versioned transcrypt configuration. 

### Salt Method

The current salt method has a security vulnerability. This is because it is partially derived from the password. A clever attacker can take advantage of this to perform a low-cost brute force attack on the password. It might be better to not writeup the details of how to perform the attack on a public page.

Ultimately the problem is that the password is used to determine the password along with the file name that is being encrypted and a hash of its contents. So how do we fix it? We need the salt to be deterministic so git doesn't think the file changed when it hasn't. 

How can we do this? We could choose a random password when transcrypt is initialized and then save that to an unencrypted config file. If we check that file into the repo we can ensure different checkouts also operate in the same way. The current way I've done this is with the idea of a `.transcrypt/config` files that stores the values `trancrypt.config-salt` (this new file is managed with git's configuration tool, which is very versatile and perfect for this task).

[...]

PR #136, opened 2022-06-14
Alternative enhancement

Erotemic:
This is an alternative to #132 that aims to be simpler. The new CLI args are:

```
--digest=<openssl-digest> 
--salt-method=<'password' | 'random' | custom_string>
--kdf=<"none" | "pbkdf2">
```

By default these will always write a new `.transcrypt/config` file that is then `git add`-ed to the repo. In my own experience it is a pain to have to re-enter these non-secret data that could instead be stored in plaintext in the repo. We should also make an option to optionally disable it, but I think 99% of users benefit.

The priority of loading a variable is: 

1. CLI args
2. unversioned .git/config
3. versioned .transcrypt/config
4. user prompt or crash if non-interactive

The exception is `transcrypt.password` and `transcrypt.openss-path` which are never saved or loaded from the `.transcrypt/config` file (only the unversioned `.git/config` file). The other behavior of note is that when `--salt=random`, the program will replace it with a random value before the main logic starts.

@elasticdog @jmurty The other posts and writeup I've made haven't generated much discussion. I think enhancements along this line are a clear win, and I'd like to start recommending transcrypt to people, so it would be helpful if I knew what it would take to get this merged here. Otherwise I can continue using my fork. 

This also adds:

* environ: `TRANSCRYPT_TRACE` - if set to 1, will `set -x` to help with debugging
* environ: `EDITABLE_INSTALL` - if set to 1, will symlink transcrypt into the git repo instead of copying it. Also helpful for debugging. 
* Used bpkg conventions to allow the script to be sourced as a library. Again, this is a debugging QOL improvement, and does not impact user functionality.


jmurty:
Thanks for the simplified PR. I'm aiming to get a new 2.2.0 release with the fairly urgent OpenSSL v3 fix out to clear the way for this change which I think may be large enough to warrant a `3` major release, especially since it will break backwards compatibility for the new, improved, and stronger encryption which we should probably recommend by default.

The main block on getting this merged is for myself or @elasticdog to have sufficient time available. I can't promise when this will happen, but after we've cut the 2.2.0 release I will be willing to merge to the `main` branch and iterate on things there, rather than leave your work marooned on a PR branch.

Thanks for your patience; we'll get there.

jmurty:
Hi @Erotemic I have updated this branch with changes, mostly focussed on getting the existing BATS tests to pass to give us a solid base to work from. I haven't quite managed it yet, but it's getting close

Erotemic:
I do have a bit more work that needs to be done on this branch. Specifically, I need to rename the arguments used in the program to the arguments I've described above.

I agree that getting 2.2.0 out with the SALTED changes is by far the most important thing. Keep me updated with status on that.

jmurty:
I think a rekey should regenerate a new `base-salt`. We might need to give the user a way to keep the old base salt if they want to, but this wouldn't be the default.

Overall I still don't see the point of prompting for or storing a salt **method**. Is there a scenario not covered in the table below where we need to know the salting method?

| Using PBK?  | Base salt in config? | Outcome |
| ------------- | ------------- | ------------- |
| Yes  | Yes  | OK |
| Yes |  No | Error |
| No |  Yes | Ignored |
| No |  No | Ignored |

Erotemic:
Yes: the case where the user wants to keep their base salt on a rekey. We need to know if the base-salt was generated via a random process or if the user explicitly set it to that. But we could always not support that case. It does weaken the security.

jmurty:
We should definitely discourage keeping the same base salt on a rekey. In this case I don't think we need to know whether an existing base salt was randomly or explicitly set, just whether the user wants to keep it.

For an interactive rekey we could prompt to keep the existing base salt – with No as the default – then let them enter a new base salt or leave blank to generate new a random one.

For a non-interactive rekey, accept a new argument like `--base-salt` with the base salt value to keep, otherwise generate a new random one.

jmurty:
Though I'm very tempted to always regenerate the base salt on a rekey, and add preservation of the base salt only if/when someone actually asks for it. As you say it weakens security, and it complicates the UI and our code for an unlikely and ill-advised use-case.

Erotemic:
I got a little bit of time to work on this, and it seems to be coming together. I rebased on main and fixed the conflicts.

In working on, I found the case table needed to be expanded to include what action was taking place: Are we running a helper-script (e.g. smudge, clean), configuring transcrypt, or re-keying transcrypt? 

The helper-script cases are as you described:

```
    +---------------+-------------+--------------------+------------------+
    | Action        | Using_KDF   | BaseSaltInConfig   | Outcome          |
    |---------------+-------------+--------------------+------------------|
    | helper-script | True        | True               | Use Config Value |
    | helper-script | True        | False              | Error            |
    | helper-script | False       | True               | Ignored          |
    | helper-script | False       | False              | Ignored          |
    +---------------+-------------+--------------------+------------------+
```

But for configure and rekey we need to additionally know if the user specified the base-salt on the command line. In that case, here are the cases:

```
    +-----------+-------------+------------------+--------------------+------------------------+
    | Action    | Using_KDF   | BaseSaltInArgv   | BaseSaltInConfig   | Outcome                |
    |-----------+-------------+------------------+--------------------+------------------------|
    | configure | True        | True             | True               | Use Argv and Overwrite |
    | configure | True        | True             | False              | Use Argv and Overwrite |
    | configure | True        | False            | True               | Use Config Value       |
    | configure | True        | False            | False              | Randomize              |
    | configure | False       | True             | True               | Ignored                |
    | configure | False       | True             | False              | Ignored                |
    | configure | False       | False            | True               | Ignored                |
    | configure | False       | False            | False              | Ignored                |
    +-----------+-------------+------------------+--------------------+------------------------+
    +----------+-------------+------------------+--------------------+------------------------+
    | Action   | Using_KDF   | BaseSaltInArgv   | BaseSaltInConfig   | Outcome                |
    |----------+-------------+------------------+--------------------+------------------------|
    | rekey    | True        | True             | True               | Use Argv and Overwrite |
    | rekey    | True        | True             | False              | Use Argv and Overwrite |
    | rekey    | True        | False            | True               | Randomize              |
    | rekey    | True        | False            | False              | Randomize              |
    | rekey    | False       | True             | True               | Ignored                |
    | rekey    | False       | True             | False              | Ignored                |
    | rekey    | False       | False            | True               | Ignored                |
    | rekey    | False       | False            | False              | Ignored                |
    +----------+-------------+------------------+--------------------+------------------------+
```

I think sits on a really nice tradeoff point between configurability and security. The user is now never prompted for the base salt. Most users wont care, and the ones that do can specify it on the command line in either the configure or rekey step. Otherwise we always randomize on a rekey and on initial configuration.

The last piece to this puzzle is to re-enable the versioned config and to give the user control over if they want to use it or not. To this end I added another command line parameter: `--versioned-config`. If the user sets this to 0 or 1 it will force it to be disabled/enabled regardless of any other settings. The final setting of "use-versioned-config" will be written to the unversioned config at configure / rekey time, which will allow the helper script to know if they should or should not use it.

The default setting depends on if the user specified that argument, if a `.transcrypt/config` versioned config file already exists, and if the configuration is using KDF or not. The truth table for the defaults I settled on are:

```
+-------------------------+---------------------------+-------------+------------------------+
| argv_versioned_config   | versioned_config_exists   | Using_KDF   |   use-versioned-config |
|-------------------------+---------------------------+-------------+------------------------|
| 1                       | True                      | True        |                      1 |
| 1                       | True                      | False       |                      1 |
| 1                       | False                     | True        |                      1 |
| 1                       | False                     | False       |                      1 |
| 0                       | True                      | True        |                      0 |
| 0                       | True                      | False       |                      0 |
| 0                       | False                     | True        |                      0 |
| 0                       | False                     | False       |                      0 |
| ''                      | True                      | True        |                      1 |
| ''                      | True                      | False       |                      1 |
| ''                      | False                     | True        |                      1 |
| ''                      | False                     | False       |                      0 |
+-------------------------+---------------------------+-------------+------------------------+
```

I've added code to the Python tests to verify these new configurations work as expected. I've also updated the algo writeup, man pages, and completion scripts (not 100% sure I did zsh correctly, but the bash one seems to work)

jmurty:
Hi @Erotemic I have just released version [2.2.0](https://github.com/elasticdog/transcrypt/releases/tag/v2.2.0) of Transcrypt, so the main branch is now clearer to prepare the improvements in this branch for merging.

I'd like to focus on the KDF/PBKDF2 key derivation function (and base salt) first, since these are the most important changes. I'd like to focus on the usability of the new arguments, add some explanations for what the settings mean at important points (e.g. the interactive config step), and make sure it's as easy as possible to use transcrypt in the new and best way: with PBKDF2 and SHA256.

Some initial thoughts/questions:
- Could we reduce the size of the generated random base salt? 32 hex characters is a lot, would 16 do (like OpenSSL's own salt value)
- Can we detect whether the user's `openssl` version is compatible with KDFs, and which ones exactly? The command `openssl list -kdf-algorithms` seems to work
  - Once we can detect KDF compatibility, we can show a useful error message if the user tries to configure a KDF and/or prompt the user to install a newer OpenSSL version for better security
- The `--kdf` argument should accept any valid KDF value – not just `pbkdf2` – or if Transcrypt will only work with `pbkdf2` we might as well only add the `-pbkdf2` shortcut argument
  - As such, I'd like to change the values accepted by `--kdf` from `0`, `1`, `none`, `pbkdf2` to any valid KDF value
  - Instead of accepting `--kdf=none` could we interpret the absence of the `--kdf` argument as none?
- I'd like to make the default digest and KDF settings `sha256` and `pbkdf2` respectively, at least for the interactive configuration process. Defaults are important, and since we're allowing for better security with these changes we should strongly encourage users to do what's best for them.

By the way, I don't expect you to make all these changes, I'll work on them too. This list ☝️ is mainly a brain-dump of things I plan to start on.

Although I think it makes sense to keep the versioned config feature in this branch alongside the KDF changes, I expect I will merge just the KDF changes first, then follow up on the versioned config feature in a second pass.

Erotemic:
### On Versioned Configs

So your thought is that we will store the base salt and everything else locally without the versioned config, which adds in all of those features, but just forces the user to copy-paste rather long commands to initialize new clones (as there is no other mechanism for passing around the salt value)? I think that's fine in terms of splitting up the PR, but I just want to make sure you're not thinking about releasing transcrypt without the option for a versioned config.

### Ok KDFs

I like removing -pbkdf2 and using -k. 

~From what I understand openssl doesn't support anything except pbkdf2. It's unfortunate. It would have been nice to use something like Argon2 or scrypt. I'm all for removing 1 and 0, they were really placeholders anyway to make getting the proof-of-concept out easier.~ Turns out I'm wrong. `openssl list -kdf-algorithms` does this (looks like a pain & a half to parse though). We should absolute support things beyond pbkdf2, up until now I thought it was the only one. Looks like scrypt is available, but unfortunately argon2 is not in 3.0.2. Maybe soon. 

I strongly recommend against removing kdf=none. I'm ok changing it to `--kdf=''` as an empty string though. Having an explicit value that represents the absence of a KDF  is important to me. One of my biggest pet peeves is when I can't specify every argument to a CLI explicitly. It makes it more tedious to use things programatically. I'm a big believer that CLIs should have a key/value mechanism for specifying all ways they could be parameterized. This makes it easy to represent sets of arguments as a dictionary, which I use in my tests. I'm just a contributor here, so I'll accept pushback on this and deal with the annoyance, but I hope this mini-rant is enough to convince you that having an explicit way to specify all key/value pairs of arguments on the CLI is a desirable thing.

Also, using the absence of `--kdf` on the command line doesn't work when pbkdf2 is the default (how do you specify legacy behavior?)

### On Salt

I don't have a problem with using 16 instead of 32 characters. It was an arbitrary choice, and I don't think it is a significant reduction in any security. We might want to bump it up a little from 16, maybe to 20 or 24, because unlike openssl we aren't choosing a random salt each time, we are reusing it. It might make sense to push up the entropy here just in case. I don't have any theoretical justification for this, it's just a feel-fact.

### On Defaults

I'm all for using sha256 and pbkdf2 as defaults... but I'm not sure about the UX for people already using legacy settings. If they upgrade transcrypt and try to initialize their legacy repo does that work? The salt change isn't a big deal, although every file in the repo will show up as having a diff, but the kdf change will not correctly decrypt the data unless the user who is upgrading realizes that their has been a change they need to account for (which 9 / 10 times they will not).

I think we need to write a few test cases that test how the program works through the upgrade process from 2.x to 3.x. I don't have those in my python test suite right now, but I can put some effort into adding them. Not sure how easy it would be to add those cases to the bats suite.

My thought is that the first release of 3.0 should contain all of these options, but not modify any defaults. It could emit warnings to user to give them a heads up that they are using legacy settings as well as instruction on how to upgrade them and that defaults will change in the future. I think a 3.1 release could then switch the defaults with minimal disruption. 

Note, if we _do_ release a version with kdf, digest, and salt _without_ the versioned config, then we should certainly not change the defaults in that case. But once the versioned config is available, I think there is a path for sensible default upgrades.

jmurty:
Quick clarifications:
- I don’t plan to release without the versioned config option – that’s helpful – but first up I’d like to work through the UX without relying on that config help so it is as usable and clean without it as possible
- I agree the pasting the KDF listing looks like a massive pain. Maybe we can just hard-code pbkdf2 and scrypt for now
- I’d prefer `--no-kdf` or similar to `--kdf=none`. I haven’t come across other tools that use “none” or “0” in this way, and since we’re built on Git following it’s argument conventions seems reasonable
- Making more secure options the default will be tricky for backwards compatibility. We can probably only make them the default (recommendations) in the interactive config step, but even that would be a win
- I also don’t have an informed opinion on the best length for base salt. The long value is only really a problem for sharing configs as command + args only, without the versioned config, but it does make that command more unwieldy. I'll need to give this some more thought, and maybe research

Erotemic:
So, it doesn't seem to be straightforward to switch between the underlying kdfs. `openssl enc` supports the `-pbkdf2` option, which just enables that particular one, to my earlier rant, there does not seem to be a key/value mechanism that allows a user to simply specify which one you want to use on the command line. This might involve looking into the approach you suggested earlier where we run the KDF once and then store the key and IV and use that directly (which would also give a nice speed boost for diff and status operations).

It would be nice to get an opinion about the salt length from someone who really knows what they are doing wrt to cryptography. I know it's best to use a unique salt for each file --- and we effectively do by mixing in the hash of the file --- but I don't have a sense for if lengthening the salt provides any extra security that could make up for the fact that there is this public content that contributes to the security of all data in the repo. I also think we shouldn't worry too much about it. Using a salt of length 16 is probably fine, and there is nothing stopping the user from specifying a custom base-salt that is longer. It might be best to just change it to 16 and leave this as a known open question.

#### Again on the key/value

I'm going to argue again in favor of reserving _some value_ for kdf that indicates it is not used (I don't care what it is).  The  "--option" and "--no-option" makes sense in the case where option takes on a binary value. For things that could take on more than one value (we would like to be able to say `--kdf=pbkdf2` and `--kdf=scrypt`), using a key/value CLI specification makes the most sense. This pattern does exist in git, especially in the `git config` command, where everything is key/value based. I'm a python guy so `--kdf=none` makes a lot of sense to me, but `--kdf=null` would also be fine. I'd be happy with any of the following `--kdf=` ... : `none`, `off`, `disabled`, `0`, `null`, etc... 

(Note: previously I said  `--kdf=''` would be ok, but after thinking about it  I realize this is actually a bad idea because it becomes ambiguous with the case where the user wants no kdf versus the user didn't specify what they wanted). 

I could live with `--no-kdf`, but I'd strongly recommend against it. The reason is that if you are scripting something its a lot easier when one part of the argument stays the same in all cases, adding the `--no-` convention means anyone who is scripting on top of this will need to have a special case to fixup the command line when they need to disable the feature, whereas every other change would just require changing the right-hand-side of the `--kdf=<rhs>` argument.

Erotemic:
I want to note that I've been using this in my day-to-day, and it does seem that there is an issue with it that should be resolved before merging.

When I checkout different branches, even though they were encrypted with this branch, it complains that there would be conflicts in the encrypted files (even though in the plaintext version of them there isn't). That means that there is some non-deterministic or different thing happening either between machines (where perhaps I have different version of openssl installed) or on the same machine. I haven't localized it yet. Just noting it here for reference.

jmurty:
Hi @Erotemic thanks for noting those cross-machine issues. Can you note down anything that might help to debug these kinds of problems, like the platform and OpenSSL versions on the different machines?

Erotemic:
They were both Ubuntu machines. I think one is 22.04 and the other is 20.04, so they come with their respective versions of openssl. Should be fairly simple to fire up a few docker containers to test this.

Erotemic:
I made a small proof-of-concept file: tests/test_multi_docker.py

This ports the docker-management file `oci_container.py` from [cibuildwheel](https://github.com/pypa/cibuildwheel/blob/main/cibuildwheel/oci_container.py) which lets me start two docker images in a single python session and then send commands to each. In the example I set them both up with a minimal transcrypt repo and I try to perform some actions, commit on one, and then pull on the other.

I think this is exposing some issues in this branch, and I'll have to nail those down... eventually.

jmurty:
In excellent news, it looks like the latest version of MacOS 13 Ventura includes a newer version of LibreSSL (3.3.6) with an `openssl` command that supports PBKDF2:

Using the built-in MacOS version of `openssl` from LibreSSL:
```bash
$ /usr/bin/openssl version
LibreSSL 3.3.6

# Encrypt with PBKDF2, 2500 iterations
$ echo "TEST" | ENC_PASS=abc123 /usr/bin/openssl enc -e -a -aes-256-cbc -md sha256 \
  -S deadbeef00000000 -pass env:ENC_PASS -pbkdf2 -iter 2500
U2FsdGVkX1/erb7vAAAAAN4Q77gUNYGeQo3tucwenVM=

# Decrypt with PBKDF2, 2500 iterations
$ echo U2FsdGVkX1/erb7vAAAAAN4Q77gUNYGeQo3tucwenVM= \
  | ENC_PASS=abc123 /usr/bin/openssl enc -d -a -aes-256-cbc -md sha256 \
  -S deadbeef00000000 -pass env:ENC_PASS -pbkdf2 -iter 2500
TEST
```

The MacOS / LibreSSL version is also compatible with OpenSSL version 1.1:
```bash
$ echo "TEST" | ENC_PASS=abc123 /opt/homebrew/opt/openssl@1.1/bin/openssl enc -e -a -aes-256-cbc -md sha256 \
  -S deadbeef00000000 -pass env:ENC_PASS -pbkdf2 -iter 2500
U2FsdGVkX1/erb7vAAAAAN4Q77gUNYGeQo3tucwenVM=
```

And with OpenSSL version 3, once our (ugly) work-around is applied to include the salted prefix that version 3 stopped including (see #133):
```bash
# OpenSSL v3+ without salt prefix DOES NOT MATCH
$ echo "TEST" | ENC_PASS=abc123 /opt/homebrew/opt/openssl@3/bin/openssl enc -e -a -aes-256-cbc -md sha256 \
  -S deadbeef00000000 -pass env:ENC_PASS -pbkdf2 -iter 2500
3hDvuBQ1gZ5Cje25zB6dUw==

# OpenSSL v3+ without salt prefix work-around does match
$ (printf "Salted__" && printf "%s" "deadbeef00000000" | xxd -r -p && echo "TEST" | ENC_PASS=abc123 /opt/homebrew/opt/openssl@3/bin/openssl enc -e -aes-256-cbc -md sha256 -S deadbeef00000000 -pass env:ENC_PASS -pbkdf2 -iter 2500) | /opt/homebrew/opt/openssl@3/bin/openssl base64
U2FsdGVkX1/erb7vAAAAAN4Q77gUNYGeQo3tucwenVM=
```

Annoyingly our work-around for OpenSSL v3 no longer including the salt prefix is probably broken on MacOS Ventura, because LibreSSL's `openssl` command reports a major version of `3` but **still includes the salt prefix** while we forcibly add the salt prefix for major version 3. It's good that LibreSSL doesn't make breaking changes as freely as OpenSSL does, but it means we need to be smarter in applying our work-around for OpenSSL v3.

jmurty:
As an example of the kinds of compatibility problems we're facing that I worried about in https://github.com/elasticdog/transcrypt/issues/134#issuecomment-1475209945, I just wasted some time trying to get the BATS tests to pass for this PR's code using the LibreSSL 3+ `openssl` command included with MacOS 13.

The stumbling block was in trying to get a list of available digests from LibreSSL, which:
- does not support listing available digests with either of the `list-digest-commands` or `list -digest-commands` used by different versions of OpenSSL, and instead requires `list-message-digest-commands`
- returns a zero error code (no error) when you use one an incorrect command, so we can't easily tell that the command failed.

I'm not sure how best to work around the faulty error code result to detect failed commands?
- check for a zero-length stderr stream after the command
- look for the word "invalid" in the output since both OpenSSL and LibreSSL happen to use this word (though with different capitalisation)
- some other much more sensible option?

jmurty:
Hi @Erotemic I just created (yet another) PR for the PBKDF2 work: #162

I have made some time recently to push the PBKDF2 improvements ahead, and thanks to your work here I have made a lot of progress. I started a new PR instead of building on this one so I could build things up slowly, keep everything in my head as I went, and make the smallest number of changes possible to get things working. I have been lifting your code pretty liberally across to the new PR, but only the pieces I need as I need them.

A key difference in the new PR is how the base salt is handled. I _think_ I've come up with an easier and strong-enough way of doing this so we don't need to store or prompt for a long salt value: the base (called "project" in the new PR) salt is derived from the password using the same PBKDF2 settings as all other files, to make the partial hash value used in per-file salts just as difficult to brute-force as any other file in the repo.

I describe this in more detail in the PR comment section _Deterministic salting for PBKDF2 (Feedback needed)_. I'd really appreciate your thoughts on this approach.

Erotemic:
Stale PR. New work should happen elsewhere.

PR #137, opened 2022-06-15
Use core attributesFile from worktree

dimitrov-adrian:
When using attribute file overrides (for example if override in .dotfiles or by other includes), get the core.attributesFile from the current worktree if available.


jmurty:
Thanks for the improvement!

Issue #141, opened 2022-07-08
Consider making functions for get_password() and set_password()


ljm42:
I would like to move the transcrypt password out of the git config file and come up with some other method of storing it. 

I'm wondering if you'd be open to creating functions for `get_password()` and `set_password()`, something like this:

```
get_password() {
  local password=$(git config --get --local transcrypt.password)
  echo "$password"
}

set_password() {
  local password=$1
  git config transcrypt.password "$password"
}
```
the code would call those like this:
```
  current_password=$(get_password)
  set_password "$password"
```
Then in my implementation I would just have to modify those functions and not all the places that get/set the password. Would you be open to that?

Oh, with these names the current `get_password()` would have to be renamed, maybe `get_password_from_user()` ?

jmurty:
Solved with new `load_password()` and `save_password()` functions merged from #142, thanks for the improvement!

PR #142, opened 2022-07-09
create functions for get_password() and set_password()

ljm42:
This makes it easier to override password handling locally
See #141

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -25,7 +25,7 @@ readonly DEFAULT_CIPHER='aes-256-cbc'
 
 # get encryption password
 # by default is stored in git config, modify this function to move elsewhere
-get_password() {
+load_password() {
 	local password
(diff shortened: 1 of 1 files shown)


jmurty:
Hi @ljm42 I like this change thanks, but I think it could be even clearer if the new functions could be called `load_password()` and `save_password()` instead of using the `get`/`set` prefixes. This would also remove the need to rename the existing `get_password()` function to `get_password_from_user()`

What do you think?

ljm42:
Great idea! Can you squash these to a single commit? Or I could create a new PR on a clean branch if you'd like?

jmurty:
It’s fine to make changes on this branch. I’ll squash at merge time

ljm42:
Perfect. This should be good unless you see other issues

PR #157, opened 2023-02-13
be verbose when you say the repo is dirty

andreineculau:
useful for those less knowledged or in CI flows

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -429,6 +429,7 @@ run_safety_checks() {
 		git update-index -q --really-refresh
 		# check if the repo is dirty
 		if ! git diff-index --quiet HEAD --; then
+			git diff-index HEAD --
 			die 1 'the repo is dirty; commit or stash your changes before running transcrypt'
 		fi
 	fi


jmurty:
I like the idea here @andreineculau but the output from `git diff-index HEAD --` isn't very user friendly, e.g:
```
:100755 100755 8217bfddbf441278e55c1416eca90a6d7ea3afe4 0000000000000000000000000000000000000000 M      transcrypt
```

I'm trying some other options to get nicer formatting

jmurty:
I've tweaked your approach to print an easier to understand listing of dirty files, e.g:
```
transcrypt: the repo is dirty; commit or stash your changes before running transcrypt

M       CHANGELOG.md
```

Thanks for the inspiration @andreineculau!

andreineculau:
Great stuff! Thanks!

I don't remember the background but I believe the "nicer" printouts might be hiding info (corner-cases), that's why I went with the basix printout. This is 2? years ago, git has improved, I don't remember the details 🤷

PR #167, opened 2023-05-25
Fix #166

### LINKED ISSUE #166: Cannot detect transcrypt repo
Get error `transcrypt: the current repository is not configured` using `transcrypt --display` command

It seems that it caused when `grep` is set globally to use colors, then the comparison from get_contexts_from_git_config() the `[[ "$name" = "transcrypt.password" ]]` do not work since it compare colorized than non-colorized string

dimitrov-adrian:
Disable colors when doing grep

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -1041,7 +1041,7 @@ import_gpg() {
 
 # Echo space-delimited list of context names defined in the git config
 get_contexts_from_git_config() {
-	config_names=$(git config --local --name-only --list | grep transcrypt)
+	config_names=$(git config --local --name-only --list | grep --color=never transcrypt)
 	extract_context_name_regex="transcrypt\.(${CONTEXT_REGEX}).password"
 	contexts=()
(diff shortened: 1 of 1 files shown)


jmurty:
Hi @dimitrov-adrian thanks for this PR but I'm reluctant to add `--color=never` to grep commands because although that flag works for GNU GREP, IT may not be supported on all platforms and could break for some users. Also, if the `grep` command needs to have `--color=never` added in this place, there are a few other uses where it will have the same problem.

Is there a safer way to disable grep colour output that doesn't assume the `--color` flag is always supported, perhaps by setting an environment variable?

If we can find a risk-free way of disabling color output I'd be happy to merge that, but the best fix might be to change the way you have grep colouring set up per my update on issue #166

jmurty:
Fixed with alternative approach in commit a825f9de0b16084b5e279cf9deaba23ade333b85

Issue #173, opened 2023-10-14
Handle quotes and colon in filenames


andreineculau:
Somewhere we don't escape filenames properly, ending up with errors like

```
fatal: ambiguous argument ':"Workspace/ETL/Jobb/Generalist Marketplace/ismdou - Know the identifiers of the table \"FOO\".py"': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
fatal: path 'Workspace/ETL/Jobb/Generalist Marketplace/ismdou - Service Finder Analysis (stakeholder ' does not exist (neither on disk nor in the index)
fatal: path 'Workspace/Users/username@example.com/Accelerators/Wide&Deep Recommender/MR 01' does not exist (neither on disk nor in the index)
fatal: path 'Workspace/Users/username@example.com/Accelerators/Wide&Deep Recommender/MR 02' does not exist (neither on disk nor in the index)
fatal: path 'Workspace/Users/username@example.com/Accelerators/Wide&Deep Recommender/MR 03' does not exist (neither on disk nor in the index)
```

The first fatal is due to double quotes in the filename. The rest are due to colon in the filename (not visible, but right after the path in the error message; filename is truncated in the error message).

PS: The worst of it is that the commit goes through, so you end up with a commit that should have encrypted files, but instead files are in plain text.

jmurty:
Hi, I have done some testing and the problem is triggered by the presence of quote characters (`"`) in filenames, how these quotes are output by Git commands and then fed into follow-on commands within Transcrypt.

At multiple places in the Transcrypt script we use a command like this to find encrypted files (tracked files to which the `crypt` filter is applied):

```bash
git -c core.quotePath=false ls-files | git -c core.quotePath=false check-attr --stdin filter | awk 'BEGIN { FS = ":" }; /crypt$/{ print $1 }'
```

These commands try to avoid problems with special/unusual characters in filenames by disabling Git's `core.quotePath` config setting, but unfortunately per [the documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corequotePath):
> Double-quotes, backslash and control characters are always escaped regardless of the setting of this variable.

Running the above command on a repo containing files with quote characters produces output like this:
```
"ismdou - Know the identifiers of the table \"FOO\".py.secret"
sensitive_file
```

It is because the ismdou… file gets quoted and quote-escaped in this way that follow-up commands fail. 

[...]

PR #174, opened 2023-10-22
Fix handling of double-quotes in encrypted file names #173

jmurty:
See issue #173


andreineculau:
Nice patch! This fixes also another bug where `transcrypt --list` would list filenames containing the word `crypt`, rather than having the filter `crypt`.

andreineculau (approved):

ctapobep:
Does this also fix the names with the whitespaces? Will it be released soon? It's been a year..

PR #186, opened 2024-11-05
better support for git worktree

andreineculau:
running `git worktree add some-branch some-branch` in a transcrypted repository fails because the "common" config will point to a .git/worktrees/some-branch/transcrypt instead of .git/worktrees/transcrypt

this fix is half-baked because it then makes it impossible to activate transcrypt only inside 1 worktree, but overall it makes sense

### DIFF
--- a/tests/test_contexts.bats
+++ b/tests/test_contexts.bats
@@ -66,10 +66,10 @@ function teardown {
 
   # Use --git-common-dir if available (Git post Nov 2014) otherwise --git-dir
   # shellcheck disable=SC2016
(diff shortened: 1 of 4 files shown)


jmurty:
Thanks for this, I have extended your change so transcrypt (and tests) uses the `--git-common-dir` option instead of `--git-dir` in all places. 

Does this change make sense? And does it get transcrypt closer to working properly with worktrees (with the caveat you mentioned)?

andreineculau:
Looks good 👍

johnjensenish:
Hey @jmurty I've done testing of this and the worktree support has been great in my workflows.

I would deploy this change widely in our dev org if there were a new public release, so that we don't need to worry about the `-pre` suffix in our tooling.

jmurty:
Thanks @andreineculau for the fix, and everyone else for the feedback and the push to release. The improved `git worktree` support is now available in release [transcrypt 2.3.2](https://github.com/elasticdog/transcrypt/releases/tag/v2.3.2)

PR #193, opened 2025-01-30
performance: don't create subshells

andreineculau:
I have some repos with many many files.
I haven't checked the git history for a reason on using a for loop and subshell and bash pattern checking instead, but going with piping and grep and sed makes a night and day difference.

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -173,20 +173,25 @@ _list_encrypted_files() {
 	# List files with -z option to disable quoting of filenames, then
 	# immediately convert NUL-delimited filenames to be newline-delimited to be
 	# compatibility with bash variables
-	for file in $(git ls-files -z | tr '\0' '\n'); do
-		# Check for the suffix ': filter: crypt' that identifies encrypted file
-		local check
-		check=$(git check-attr filter "$file" 2>/dev/null)
-
-		# Only output names of encrypted files matching the context, either
-		# strictly (if $1 = "true") or loosely (if $1 is false or unset)
-		if [[ "$strict_context" == "true" ]] &&
-			[[ "$check" == *": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}" ]]; then
-			echo "$file"
-		elif [[ "$check" == *": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}"* ]]; then
-			echo "$file"
-		fi
-	done
+	# List files with -z option to disable quoting of filenames,
+	# then filter for files marked for encryption (git check-attr + grep),
+	# then only keep filenames (sed)
+	# then evaluate escaped characters like double-quotes, backslash and control characters (eval)
+	# which are part of the output regardless of core.quotePath=false as per
+	# https://git-scm.com/docs/git-config#Documentation/git-config.txt-corequotePath
+	git -c core.quotePath=false ls-files -z | tr '\0' '\n' |
+		git -c core.quotePath=false check-attr filter --stdin 2>/dev/null |
+		{
+			# Only output names of encrypted files matching the context, either
+			# strictly (if $1 = "true") or loosely (if $1 is false or unset)
+			if [[ "$strict_context" == "true" ]]; then
+				grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}$" || true
+			else
+				grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*$" || true
+			fi
+		} |
+		sed "s|: filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*||" |
+		while read -r file; do eval "echo $file"; done
 }
 
 # Detect OpenSSL major version 3 or later which requires a compatibility
@@ -1035,7 +1040,14 @@ upgrade_transcrypt() {
 list_files() {
 	if [[ $IS_BARE == 'false' ]]; then
 		cd "$REPO" >/dev/null || die 1 'could not change into the "%s" directory' "$REPO"
-		_list_encrypted_files true
+
+		if [[ -z "$CONTEXT_CRYPT_SUFFIX" ]]; then
+			# Non-strict listing of files when context is unset / default
+			_list_encrypted_files
+		else
+			# Strict listing of files when a specific context is set
+			_list_encrypted_files true
+		fi
 	fi
 }
 


andreineculau:
Git blame took me to https://github.com/elasticdog/transcrypt/commit/6e557df8629967397f83e72944c8fc0d7d3ba7e7 which fixed support for double-quoted filenames...

transcrypt
```diff
@@ -173,20 +173,15 @@ _list_encrypted_files() {
...
-	done
+    git ls-files -z | tr '\0' '\n' \
+      | git check-attr filter --stdin 2>/dev/null \
+      | {
+      if [[ "$strict_context" == "true" ]]; then
+        grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}$" || true
+      else
+        grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*$" || true
+      fi; } \
+      | sed "s|: filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*||"
```
andreineculau, line 184: lint

```suggestion
	git ls-files -z | tr '\0' '\n' |
		git check-attr filter --stdin 2>/dev/null |
		{
		if [[ "$strict_context" == "true" ]]; then
			grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}$" || true
		else
			grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*$" || true
		fi; } |
		sed "s|: filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*||"
```

andreineculau:
TDD FTW 🚀  Failing on

```
not ok 33 crypt: handle challenging file names when 'core.quotePath=true'
```

Will have a look and see what could be the alternative

transcrypt
```diff
@@ -173,20 +173,15 @@ _list_encrypted_files() {
...
-	done
+	git ls-files -z | tr '\0' '\n' |
+		git check-attr filter --stdin 2>/dev/null |
+		{
+		if [[ "$strict_context" == "true" ]]; then
+			grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}$" || true
+		else
+			grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*$" || true
+		fi; } |
+		sed "s|: filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*||"
```
andreineculau, line 184:
```suggestion
		sed "s|: filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*||" |
		while read -r file; do eval "echo $file"; done
```

transcrypt
```diff
@@ -173,20 +173,16 @@ _list_encrypted_files() {
...
-		fi
-	done
+	git ls-files -z | tr '\0' '\n' |
+		git check-attr filter --stdin 2>/dev/null |
+		{
+		if [[ "$strict_context" == "true" ]]; then
+			grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}$" || true
+		else
+			grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*$" || true
+		fi; } |
```
andreineculau, line 183:
```suggestion
			if [[ "$strict_context" == "true" ]]; then
				grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}$" || true
			else
				grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*$" || true
			fi;
		} |
```

transcrypt
```diff
@@ -173,20 +173,17 @@ _list_encrypted_files() {
...
-		fi
-	done
+	git ls-files -z | tr '\0' '\n' |
+		git check-attr filter --stdin 2>/dev/null |
+		{
+			if [[ "$strict_context" == "true" ]]; then
+				grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}$" || true
+			else
+				grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*$" || true
+			fi;
```
andreineculau, line 183:
```suggestion
			fi
```

transcrypt
```diff
@@ -173,20 +173,17 @@ _list_encrypted_files() {
...
-		# strictly (if $1 = "true") or loosely (if $1 is false or unset)
-		if [[ "$strict_context" == "true" ]] &&
-			[[ "$check" == *": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}" ]]; then
-			echo "$file"
-		elif [[ "$check" == *": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}"* ]]; then
-			echo "$file"
-		fi
-	done
+	git ls-files -z | tr '\0' '\n' |
+		git check-attr filter --stdin 2>/dev/null |
```
andreineculau, line 177:
```suggestion
		git -c core.quotePath=false check-attr filter --stdin 2>/dev/null |
```

`--core.quotePath=false` is needed so that we only quote `Double-quotes, backslash and control characters` [ref](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corequotePath)

transcrypt
```diff
@@ -173,20 +173,17 @@ _list_encrypted_files() {
...
-		# Only output names of encrypted files matching the context, either
-		# strictly (if $1 = "true") or loosely (if $1 is false or unset)
-		if [[ "$strict_context" == "true" ]] &&
-			[[ "$check" == *": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}" ]]; then
-			echo "$file"
-		elif [[ "$check" == *": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}"* ]]; then
-			echo "$file"
-		fi
-	done
+	git ls-files -z | tr '\0' '\n' |
```
andreineculau, line 176:
```suggestion
	git -c core.quotePath=false ls-files -z | tr '\0' '\n' |
```

transcrypt
```diff
@@ -173,20 +173,17 @@ _list_encrypted_files() {
 	# List files with -z option to disable quoting of filenames, then
 	# immediately convert NUL-delimited filenames to be newline-delimited to be
 	# compatibility with bash variables
-	for file in $(git ls-files -z | tr '\0' '\n'); do
-		# Check for the suffix ': filter: crypt' that identifies encrypted file
-		local check
-		check=$(git check-attr filter "$file" 2>/dev/null)
-
-		# Only output names of encrypted files matching the context, either
-		# strictly (if $1 = "true") or loosely (if $1 is false or unset)
```
jmurty, line 182: To help with maintenance in the future can you please add this comment, or similar, back into your improved version?

jmurty:
Thanks for looking at this @andreineculau, there's some dark magic in those pipe commands but if it runs much faster it's well worth the change. What kind of performance improvements are you seeing, out of interest? 

I found the test failures blocking you were because of a latent bug in my older code: the text content checks in your changes were more correct than mine, and highlighted a problem elsewhere. I have pushed a fix for the code that calls the `_list_encrypted_files()` function to only do a "strict" listing of files within a given context, when a context is actually given.

If you can add some comments, and maybe squash the PR down to a cleaner history, I should be able to merge it to `main` soon.

andreineculau:
Fantastic! I didn't have time to look into the failures and we won't have full coverage.

I will get back with some time stats and content the magic

transcrypt
```diff
@@ -173,20 +173,17 @@ _list_encrypted_files() {
...
-			[[ "$check" == *": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}" ]]; then
-			echo "$file"
-		elif [[ "$check" == *": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}"* ]]; then
-			echo "$file"
-		fi
-	done
+	git -c core.quotePath=false ls-files -z | tr '\0' '\n' |
+		git -c core.quotePath=false check-attr filter --stdin 2>/dev/null |
+		{
+			if [[ "$strict_context" == "true" ]]; then
```
andreineculau, line 179:
```suggestion
			# Only output names of encrypted files matching the context, either
			# strictly (if $1 = "true") or loosely (if $1 is false or unset)
			if [[ "$strict_context" == "true" ]]; then
```

transcrypt
```diff
@@ -173,20 +173,19 @@ _list_encrypted_files() {
 	# List files with -z option to disable quoting of filenames, then
 	# immediately convert NUL-delimited filenames to be newline-delimited to be
 	# compatibility with bash variables
```
andreineculau, line 175:
```suggestion
	# List files with -z option to disable quoting of filenames, then
	# immediately convert NUL-delimited filenames to be newline-delimited to be
	# compatibility with bash variables
	# List files with -z option to disable quoting of filenames,
	# then filter for files marked for encryption (git check-attr + grep),
	# then only keep filenames (sed)
	# then evaluate escaped characters like double-quotes, backslash and control characters (eval)
	# which are part of the output regardless of core.quotePath=false as per
	# https://git-scm.com/docs/git-config#Documentation/git-config.txt-corequotePath
```

transcrypt
```diff
@@ -173,20 +173,17 @@ _list_encrypted_files() {
...
+		git -c core.quotePath=false check-attr filter --stdin 2>/dev/null |
+		{
+			if [[ "$strict_context" == "true" ]]; then
+				grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}$" || true
+			else
+				grep ": filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*$" || true
+			fi
+		} |
+		sed "s|: filter: crypt${CONTEXT_CRYPT_SUFFIX:-}.*||" |
+		while read -r file; do eval "echo $file"; done
```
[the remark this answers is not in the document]
andreineculau: @jmurty see if https://github.com/elasticdog/transcrypt/pull/193/files#r1968164965 does the job 🚀 🙌

andreineculau:
and these are the performance metrics in a repo with 25 encrypted files, >40 000 files.

PR version, running `git commit` and thus calling transcrypt to list files

```
[master fd56d9955] foo
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foo
0.37user 0.60system 0:00.79elapsed 123%CPU (0avgtext+0avgdata 24880maxresident)k
0inputs+0outputs (408major+66302minor)pagefaults 0swaps
```

current version, running `git commit` and thus calling transcrypt to list files

```
[master f6a57465b] bar
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 bar
178.11user 198.21system 7:47.76elapsed 80%CPU (0avgtext+0avgdata 33472maxresident)k
0inputs+0outputs (381275major+74427218minor)pagefaults 0swaps
```

AND no, there is no error. It is that slow :)

jmurty:
Excellent, thanks for perfecting this huge performance improvement @andreineculau. I've merged it to `main`.

Those performance numbers from before your fix are horrifying 😱

jeancornic:
Thanks a ton for the fix @andreineculau , you really made my day.

While I am running the latest version of transcrypt (2.3.1), it took me a while to figure out that there was a fix available for my performance issue. Would it be possible to release a new version with this fix?

Issue #182, opened 2024-08-23
Test if password is correct


andreineculau:
When using the wrong password, you end up with dirty secret files.
Beyond checking for dirty secret files after "smudging", how would you check if the password/cipher is correct without touching the workspace?

jmurty:
Transcrypt isn't built to check for a correct password. To do so, it would need to track extra metadata information to detect successful or failed decryption based on the decrypted file containing or matching that metadata.

The easiest way I can think of to check for a correct password without looking at the resultant workspace files is to create a file with consistent name and known content in your repo(s), run just the smudge command directly with Transcrypt, and confirm the result is 

You would need to set the minimum required Transcrypt Git config options for this to work, but you could do it like this (using the repo's _sensitive_file_ as an example):

```bash
# Set Git config options required by transcrypt

git config transcrypt.password 'correct horse battery staple'
git config transcrypt.cipher aes-256-cbc
git config transcrypt.openssl-path openssl

# Run 'smudge' command directly. You must provide:
# - the encrypted file's contents via stdin
# - the context name (default) and the encrypted file's name via command arguments

cat sensitive_file | ./transcrypt smudge context=default sensitive_file

# Check for known content in the decrypted file

cat sensitive_file | ./transcrypt smudge context=default sensitive_file | grep -q '^I just wanna'
# $? == 0

git config transcrypt.password 'wrong password'
cat sensitive_file | ./transcrypt smudge context=default sensitive_file | grep -q '^I just wanna'
# $? == 1
```

andreineculau:
What we currently do is check how many dirty files there are before and after decrypting with transcrypt. There should be no diff, but if there is, it's because some files haven't been correctly decrypted and git sees the diff in content.

`git status --porcelain | { grep "^ M" || true; } | wc -l` gives the number of dirty files

jmurty:
That's very clever @andreineculau and much simpler than my suggestion. I'm going to build something like this into _transcrypt_ itself

jmurty:
I have cut a new release [2.3.1](https://github.com/elasticdog/transcrypt/releases/tag/v2.3.1) with an improvement to print the following message – and return a non-zero error level – when repo files are dirty after configuring _transcrypt_ which most likely means an incorrect password:

> Unexpected dirty files in the repository when configured by transcrypt, check your password

PR #195, opened 2025-02-24
Warn when password is probably incorrect with error message and return code #182

jmurty:


andreineculau:
@jmurty The reason I mentioned in #182 to compare "dirty" numbers before and after is to account for running transcrypt with `--force` (i.e. run even if dirty). If you only check dirty after, then it will always fail, even if password is correct, or?

PR #197, opened 2025-03-01
minor fixes for git_pre_commit

weichung034918:
Bash version was 5.x, this causes `git_pre_commit` to always run `slow_mode_if_failed`.
After this was fixed, empty files are also causing git_pre_commit to run `slow_mode_if_failed` again.


jmurty:
Great, thanks for this! I have merged this fix to `main`

PR #200, opened 2025-04-20
ignore global core.attributesFile

andreineculau:
scenario: you have a global .gitattributes for settings crlf, text/binary, diff etc settings

then this is what you get

```
$ transcrypt --list-contexts                                                                                                                                                             default (no patterns in .gitattributes)
custom (no patterns in .gitattributes)
```

instead of

```
$ transcrypt --list-contexts                                                                                                                                                             default
custom
```

although the .gitattributes in the repo has the patterns defined

Point is: do we really support transcrypt globally? with custom contexts?
If we do, then we need to check all attributes files, not only the first one.

PS: on a different note, it also doesn't make sense (to me at least) that `--add=<pattern>` write to anything but a local gitattributes

### DIFF
--- a/transcrypt
+++ b/transcrypt
@@ -109,7 +109,7 @@ gather_repo_metadata() {
 
 	# the current git repository's gitattributes file
 	local CORE_ATTRIBUTES
-	CORE_ATTRIBUTES=$(git config --get --local --path core.attributesFile 2>/dev/null || git config --get --path core.attributesFile 2>/dev/null || printf '')
+	CORE_ATTRIBUTES=$(git config --get --local --path core.attributesFile 2>/dev/null || printf '')
 	if [[ $CORE_ATTRIBUTES ]]; then
 		readonly GIT_ATTRIBUTES=$CORE_ATTRIBUTES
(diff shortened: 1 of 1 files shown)


jmurty:
The check against non-local `.gitattributes` was added in #137 and although I don't tend to us config files like `.gitattributes` from outside the immediate repo directory, some people do. I can imagine for example situations where you want `transcrypt` to encrypt _*.secret_ files across all your projects, in which case putting this in a global `.gitattributes` would make sense.

So I think the right fix here would be to check all attributes files, which won't necessarily be easy.

jmurty:
See a potential alternative solution that finds contexts across all gitattributes files, not just one, in #202

jmurty:
I'm sympathetic to your argument that transcrypt should expect git attributes definitions necessary for effective sharing of a repo's encrypted files to be committed to the repo as a `.gitattributes` file, rather than potentially being defined in places that aren't available to other clones of the repo.

For now I'll just fix the issue in the most straighforward way per #202, but before a new major release it would be worth taking the time to rethink and potentially clean-up this behaviour – breaking changes to simplify _transcrypt_ would be more acceptable in a major release.

PR #201, opened 2025-05-07
check for dirty files should not run on --rekey

andreineculau:
repo is expected to be dirty on `transcrypt --rekey`


jmurty:
Yes this makes sense, thanks

Issue #204, opened 2025-09-17
(Re-)Initialization of a cloned repo fails


pets-er:
Hi there!
Nice work on this tool.

I had a problem twice when trying to (re-)initialize the same cloned repo. This was not a problem on my laptops. One of the failed times was on a server that is not managed by me (work), using ssh, one of those times on termux, also using ssh.

The error that I get is this:
.git/crypt/transcrypt: eval: line 192: syntax error near unexpected token `('

It should also be noted that this leaves the repo in an encrypted state, but transcrypt thinks it is properly initialized. 

Do you have an idea how I could fix/diagnose this? As far as I know, the systems run bash (I checked using ps on the server) - the requirements should also be fulfilled. Maybe there is something odd about the setup or some implicit assumption about what the system should be capable of that is not fulfilled.
The line 192 does not seem suspicious to me - it is a comment line without a (, maybe something happening in the function above it?

Thank you!

jmurty:
I should mention that – assuming the line I linked above is indeed the problem line – you can run just the problematic function for easier testing with the `--list` param, e.g. `.git/crypt/transcrypt --list`

jmurty:
Ah okay, it definitely seems like you're hitting a transcrypt bug handling one of your file names.

Here is a command you can run which is a trimmed down version of most of what the `_list_encrypted_files()` function contains, leaving off the final problem line with the `while` loop. You can run this to bypass transcrypt entirely and debug just the commands it is running.

`IFS=$'\n' git -c core.quotePath=false ls-files -z | tr '\0' '\n' | git -c core.quotePath=false check-attr filter --stdin 2>/dev/null | { grep ": filter: crypt$" || true } | sed "s|: filter: crypt||"`

The command lists all the files in the git repo with metadata printed after the file name, then filters down to those with the metadata that indicates a transcrypt-encrypted file. You can get a sense of what is going on if you run parts of this command, building up the full thing one piped section at a time.

Does piping the above into the different variants of the `while` command I mentioned above, plus the current `while read -r file; do eval "echo $file"; done` line, reveal anything useful?

If you can share an example of the problem filename that might let me reproduce the problem locally, though I suspect it may be tricky for me to reproduce since the bug occurs for only some of your systems.

pets-er:
Okay, I checked out the issue-204 branch and initialized. This works!
Previously, I also tried the old release version, that also worked.

I have not yet tried on termux, but as the issue seemed to be the unstable version, this seems fine :)

[...]

PR #205, opened 2025-09-17
Fix listing of file names with special characters to avoid crash and unrecognised names #204

jmurty:
Tweak the file listing to fix two issues found to affect files with problematic names like `"Terrible file""")))(((][][].secret`.

This change may undo performance improvements achieved in #193.

- Ensure the `git check-attr filter` command does not quote file name characters by using `-z` arg and some extra trickery to keep space and newline delimiter characters for filter metadata and each file name respectively.

  This fixes a file named `"Terrible file""")))(((][][].secret` being listed as just `Terrible file`.

- Adjust checking for `filter` and `crypt` strings to work with changed output from `check-attr` command.

- Tweak `eval 'echo $filename'` portion of command to avoid script crash on files with special characters with error message like `eval: line 192: syntax error near unexpected token`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,8 @@ system, you must also run the `--upgrade` command in each repository:
 - Improve check for incorrect password to avoid false report when transcrypt
   init is run with --force in a repo containing dirty files & add tests (#196)
 - Greatly improve performance in a repository with many files for pre-commit
-  safety check, encrypted file listing, and showing raw file (#193)
+  safety check, encrypted file listing, and showing raw file (#193). Also fixed
+  related regression handling file names with special characters (#204)
 
 ### Fixed
 


andreineculau:
> This change may undo performance improvements achieved in https://github.com/elasticdog/transcrypt/pull/193.

confirmed 😢

andreineculau:
Thing is... I don't seem to get that error 🤔 I created a file with the mentioned filename, I staged it, and `transcrypt -l` was still happy 👀

jmurty:
Hi @andreineculau are you sure you are checking a file with a problematic name when that file is also _encrypted_, not just staged as a standard file?

I have added a unit test that demonstrates the problem with the old _transcrypt_ version from `main` (efd42e54) and proves that the fix applied on this branch solves it.

I have also made another change in 93ba18f1 which will hopefully improve performance again for repositories like yours with many files, by using an `awk` command instead of shell loop to handle the null-delimited list of filenames and their filter attributes.

Can you check the performance of this latest version of this PR?

andreineculau:
this is what I see on my end

with the new [94b8599](https://github.com/elasticdog/transcrypt/pull/205/commits/94b85993787b4646d63eb1e101687441ca07b63e) version

```
$ pbpaste > transcrypt
$ cp transcrypt .git/crypt/transcrypt
$ echo foo > "\"Difficult file name\"\"\")))(((][][].secret"
$ git add .

$ git status --porcelain
A  "\"Difficult file name\"\"\")))(((][][].secret"
M  transcrypt

$ /usr/bin/time ./transcrypt -l
"Difficult file name""")))(((][][].secret
CONST.inc.secret
        0.34 real         0.18 user         0.13 sys
```

and with the old version I have (with the changes in #193)

```
$ echo foo > "\"Difficult file name\"\"\")))(((][][].secret"
$ git add .

$ git status --porcelain
A  "\"Difficult file name\"\"\")))(((][][].secret"

$ /usr/bin/time ./transcrypt -l
CONST.inc.secret
```

👆 **NOTE** that the difficult file is not detected as encrypted!

With the version in [6aa50df](https://github.com/elasticdog/transcrypt/pull/205/commits/6aa50df95d484bb10038247f8e348b7ffd83a133) --- I just killed `./transcrypt -l` after I counted 60 seconds...

So - since we have a test that confirms the fix 🤩 and I can confirm now that the performance is on par with #193 🤩  .... 👏❤️🚀:shipit:

andreineculau (approved):

PR #209, opened 2026-05-17
Add audit and diagnostic commands

Erotemic:
With respect to my recent comment: https://github.com/elasticdog/transcrypt/issues/134#issuecomment-4468026156

This is a simple PR that is meant to help enable the transition to more secure defaults. The diagnose crypto lets the user see what sort of library support their openssl version has:

```
(uvpy3.13.13) joncrall@toothbrush:~/code/repo$ transcrypt --diagnose-crypto
Crypto diagnostics for transcrypt

  openssl path:                 openssl
  openssl version:              OpenSSL 3.0.13 30 Jan 2024 (Library: OpenSSL 3.0.13 30 Jan 2024)
  openssl enc -pbkdf2:          yes
  openssl enc -iter:            yes
  explicit -S salt behavior:    openssl-3-no-prefix
  python3 PBKDF2 fallback:      yes (Python 3.13.13)
```

And for a repo that hasn't been migrated yet, you might get something that looks like this (note this repo does have pbkdf2 enabled via my fork, it just happens to be readable via `current_crypto_setting`): 

```
(uvpy3.13.13) joncrall@toothbrush:~/code/repo$ transcrypt --audit
Security audit for this transcrypt repository

  Version:        3.0.0-pre
  Context:        default
  Cipher:         aes-256-cbc
  Crypto format:  legacy
  KDF:            pbkdf2
  Digest:         sha512
  Iterations:     256000
  Authentication: none

[HIGH] Legacy password derivation is enabled.
       Legacy transcrypt uses OpenSSL password derivation with MD5-era settings,
       which is fast to brute force for human-chosen passwords.

[HIGH] Ciphertext authentication is not enabled.
       Legacy encrypted blobs can be corrupted or tampered with without a
       strong message authentication failure before decryption.

[MEDIUM] Legacy deterministic salts are derived from the password.
         This preserves old behavior but is not suitable for a modern PBKDF2
         migration without changing the salt design.

Recommended next step: upgrade this repository to an authenticated secure format
when all collaborators have a compatible transcrypt version.
```

### DIFF
--- a/tests/test_security_audit.bats
+++ b/tests/test_security_audit.bats
@@ -1,0 +1,21 @@
+#!/usr/bin/env bats
+
+load "$BATS_TEST_DIRNAME/_test_helper.bash"
+
+@test "security: diagnose crypto reports openssl and python capabilities" {
+  run ../transcrypt --diagnose-crypto
+  [ "$status" -eq 0 ]
+  [[ "$output" = *"Crypto diagnostics for transcrypt"* ]]
+  [[ "$output" = *"openssl version:"* ]]
+  [[ "$output" = *"openssl enc -pbkdf2:"* ]]
+  [[ "$output" = *"python3 PBKDF2 fallback:"* ]]
+}
+
+@test "security: audit warns about legacy crypto settings" {
+  run ../transcrypt --audit
+  [ "$status" -eq 0 ]
+  [[ "$output" = *"Security audit for this transcrypt repository"* ]]
+  [[ "$output" = *"Crypto format:  legacy"* ]]
+  [[ "$output" = *"[HIGH] Legacy password derivation is enabled."* ]]
+  [[ "$output" = *"[HIGH] Ciphertext authentication is not enabled."* ]]
+}
--- a/transcrypt
+++ b/transcrypt
@@ -28,6 +28,14 @@ readonly VERSION='2.3.2-pre'
 # the default cipher to utilize
 readonly DEFAULT_CIPHER='aes-256-cbc'
 
+# default crypto policy for legacy repositories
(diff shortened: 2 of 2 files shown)


jmurty:
Hi, I think these new commands might be useful for us developing and debugging transcrypt as part of the PBKDF2 work, but I don't think they are appropriate to add for general use.

I think we could reduce both the diagnose and audit commands to a short warning message or two, that would recommend that users:
- rekey their repo to use PBKDF2, when that is supported by their system and not yet applied, or
- update their version of OpenSSL etc, when that is necessary to enable PBKDF2

These warnings could be shown during transcrypt init, and maybe even every time transcrypt processes files similar to the dreaded
```
```

Any transcrypt settings or setting combinations that are bad enough to prompt an audit warning should not be permitted in non-legacy repos.

jmurty:
To be clearer, I think we should detect current settings and system capabilities as these two commands are doing, but during init (to start with) not as explicit commands, and only print something like the following as a result:

When PBKDF2 is supported but legacy encryption is enabled, print something like this (imagining we have a `--strengthen` helper command for enabling PBKDF2):

    transcrypt: Upgrade this repository to use stronger PBKDF2 encryption: transcrypt --strengthen
    Warning: collaborators will need an up-to-date version of transcrypt, see <link-to-transcrypt-v3-doc>

When PBKDF2 is not supported due to an old version of OpenSSL, print something like:

    transcrypt: Stronger PBKDF2 encryption is not supported by <OpenSSL version 1.2.3>
    Please upgrade your system to enable stronger encryption, see <link-to-transcrypt-v3-doc>

PR #210, opened 2026-05-26
Prepare non secret transcrypt config

Erotemic:
This PR adds the non-secret configuration needed for the v3/PBKDF2 work without changing current behavior.

With this change transcrypt will read settings from `.transcrypt/config`, but it will only accept a configuration that matches the current v2 settings. I.e. 

```
[transcrypt]
format = legacy
cipher = aes-256-cbc
digest = MD5
kdf = legacy
```

We do not yet write .transcrypt/config on init, this PR just reads it if it is there and validates the settings. 

It **does** display the settings with `transcrypt --display`

### DIFF
--- a/docs/transcrypt-v3.md
+++ b/docs/transcrypt-v3.md
@@ -1,0 +1,28 @@
+# Transcrypt v3 development notes
+
+The v3 work introduces a repository-versioned settings file for public,
+non-secret encryption parameters. The settings file lives at:
+
+```text
+.transcrypt/config
+```
+
+The file is intended to be committed to the repository. It must not contain
+passwords or other secret material.
+
+This initial settings reader only accepts values that match the existing legacy
+encryption behavior. Later v3 changes will expand the supported settings once
+PBKDF2 encryption and decryption are implemented.
+
+Supported settings at this stage are:
+
+```ini
+[transcrypt]
+format = legacy
+cipher = aes-256-cbc
+digest = MD5
+kdf = legacy
+```
+
+Legacy repositories without this file continue to use the historical settings
+stored in local Git config.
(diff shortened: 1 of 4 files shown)


jmurty:
This is looking good thanks @Erotemic. I've applied a couple of tweaks to get the tests to pass.

Can you change the target branch to `transcrypt-v3`? I'll merge this PR to there. I want to keep the v3 work out of `main` for now in case transcrypt needs a bug fix release.

Erotemic:
Whoops, not sure how the sensitive file got modified. I was having issues with the repo state. Thanks for fixing that.

WRT to a bug fix release, there is an issue, which probably caused the sensitive file problem. I have a different default git merge config, and that - as well as the pbkdf2 output message - is messing with the smudge filters. I have the fix set aside in a standalone branch, but earlier there was discussion about not hiding the pbkdf2 warnings, and forcing the merge option for the repo seemed questionable. I wanted to think more about if it was worth merging or not, but its been over a week and I haven't come to an epiphany about it yet, so maybe I'll just push it up for discussion here? I need to prepare a MWE to demonstrate it though. Might not get to it today, but I wanted to flag that.

jmurty:
Ah interesting.

If I recall correctly, the thinking behind not hiding the pbkdf2 warning
all that time ago was that 1) it’s dishonest to hide a warning about the
non-ideal encryption, and 2) transcrypt should switch to use pbkdf2 to fix
it properly.

I didn’t manage (2) yet.

But maybe as part of the v3 pbkdf2 work we could silence that warning, and
replace it with a more actionable – and less interfering – warning from
transcrypt itself encouraging an upgrade to the newly-supported kdf-based
encryption?

Erotemic:
I force pushed this branch to squash the 3 commits so merging it wouldn't add unnecessary diff bloat.

Erotemic:
I think I may have been mistaken on the merge issue. I think the problem was I have a different global merge driver "diff3" and the test expects the default "merge", so its just that the test isn't robust to user global config, but nothing seems wrong with transcrypt itself. I also can't reproduce the issue I thought I saw with pbkdf2 causing smudge issues, so it might have been an unrelated dirty state. In any case, I'll push on the small PRs for v3 after this one lands.

PR #212, opened 2026-06-05
Fix --uninstall/--upgrade crash when saved pre-commit-crypt hook is missing

johnjensenish:
`transcrypt --uninstall` (and `--upgrade`, which calls it) crashes when a `pre-commit` hook exists but transcrypt's saved `pre-commit-crypt` copy does not:

```
transcrypt: line 905: .husky/_/pre-commit-crypt: No such file or directory
```

The read on line 905 is unguarded, so `set -euo pipefail` aborts the script mid-uninstall.

Our use of husky for pre-commit hooks makes this state likely: `core.hooksPath` points at `.husky/_`, which husky re-generates on install — removing `pre-commit-crypt` — while installing its own `pre-commit` shim. But any missing saved copy triggers it.

The fix guards the read and falls through to the existing "Cannot safely disable Git pre-commit hook" warning — without the saved copy we can't tell whether the `pre-commit` hook is transcrypt's, so leave it alone and say so.

Includes a regression test (fails on `main`, passes with the fix). Full bats suite passes; `shellcheck` and `shfmt` clean.

### DIFF
--- a/tests/test_cleanup.bats
+++ b/tests/test_cleanup.bats
@@ -99,3 +99,18 @@ SECRET_CONTENT_ENC="U2FsdGVkX1/6ilR0PmJpAyCF7iG3+k4aBwbgVd48WaQXznsg42nXbQrlWsf/
   run git show "$cached_plaintext_obj"
   [ "$status" -ne 0 ]
 }
+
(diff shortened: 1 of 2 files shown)


jmurty:
Thanks very much for this well explained and tested fix

PR #215, opened 2026-06-13
Use read, rather than awk to address NUL byte handling issues

### LINKED ISSUE #213: Homebrew installed awk is incompatible

aaronjensen:
Fixes #213

From what I can tell, accomplishes the same as the awk usage without relying on the NUL byte behavior of particular versions of awk.

I didn't add any new tests because it relies on a specific version of awk to produce the issue, but all existing tests still pass aside from one, which failed prior to my change:

```
 ✗ merge: branches with encrypted file - line changes both branches, with conflicts
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ system, you must also run the `--upgrade` command in each repository:
 
 - Show warning message instead of crashing for `--uninstall` or `--upgrade`
   action when _transcrypt_'s saved pre-commit-crypt hook file is missing (#212)
+- List encrypted files using bash's null-delimited `read` instead of
+  `awk -v RS='\x00'`, which fails to list any files with awk builds that read
+  only the first record when the record separator is a null byte, such as the
+  Homebrew "one true awk" (#213, onetrueawk/awk#270)
 
 ## [2.3.2] - 2026-05-18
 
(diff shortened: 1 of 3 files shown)


aaronjensen:
[3f24e9d](https://github.com/elasticdog/transcrypt/pull/215/commits/3f24e9dc5a5abf435bfebce42524b48a068c050c) addresses the test failure I saw

jmurty:
This is a clever fix, thanks very much!

I was concerned the different approach might cause much slower performance, so I tested the impact by comparing the old `awk` command against this new `read` approach processing a list of over 30,000 NUL-delimited file entries.

On my system the new `read` approach is about 10 times slower to process the 30k entries, but still takes only about 0.88 seconds to do so while the `awk` approach takes about 0.08 seconds. It's a shame it's slower, but I think the impact is small enough even in repos with many files that the improved compatibility is worth the change.

Merged to `main`

aaronjensen:
Ah, interesting. Performance didn’t even cross my mind, good call. Thank you for the merges and transcrypt itself.

PR #216, opened 2026-06-13
Fix symlink following for gitattributes files

aaronjensen:
If your global gitattributes file is a symlink, upgrade can fail on macOS and may replace the symlink with an actual file on Linux.


jmurty:
Thanks for the fix and test coverage. This is now merged to `main`

